PR #4, opened 2018-10-24
Fix "broken" link

smaximov:


jodosha:
@smaximov Thanks for this fix. 👍

PR #16, opened 2018-12-13
Getting Started grammar review

jacobherrington:
I did my best to improve the grammar in the guides. I intend to make another PR after this one to address what I consider style issues. 

Please let me know if any changes are necessary.

### DIFF
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -325,7 +325,7 @@ This does a lot for us:
 - Creating a view at `apps/web/views/books/index.rb` (and a spec for it),
 - Creating a template at `apps/web/templates/books/index.html.erb`.
 
-_(If you're confused by 'action' vs. 'controller,' here's a tip: Hanami only has `action` classes, so a controller is just a module to group several related actions together.)_
+_(If you're confused by 'action' vs. 'controller': Hanami only has `action` classes, so a controller is just a module to group several related actions together.)_
 
 These files are all pretty much empty.
 They have some basic code in there, so Hanami knows how to use the class.
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -121,7 +121,7 @@ This is the most popular, so you'll probably want to keep it in your future Hana
 However, there's nothing unique about this app, it's just so common that Hanami generates it for us.
 
 Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `API`, or an analytics `dashboard`.
-We could also break our `web` app into smaller.
+We could also break our `web` app into smaller parts, extracting isolated pieces of functionality.
 Hanami fully supports that, too!
 
 Different `apps` represent __delivery mechanisms__.
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -120,7 +120,7 @@ By default, we have a `web` app, which can be thought of as the standard, user-f
 This is the most popular, so you'll probably want to keep it in your future Hanami projects.
 However, there's nothing unique about this app, it's just so common that Hanami generates it for us.
 
-Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `API`, or an analytics `dashboard`.
+Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `api`, or an analytics `dashboard`.
 We could also break our `web` app into smaller parts, extracting isolated pieces of functionality.
 Hanami fully supports that, too!
 
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -472,7 +472,7 @@ $ HANAMI_ENV=test bundle exec hanami db prepare
 
 ### Working With Entities
 
-An entity is something really close to an everyday Ruby object.
+An entity is something really close to a plain Ruby object.
 We should focus on the behaviors that we want from it and only then, how to save it.
 
 For now, we need to create a simple entity class:
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -273,7 +273,7 @@ The purpose of our sample Bookshelf project is to manage books.
 
 We'll store books in our database and let the user manage them with our project.
 
-Our first step will be list out all the books we know about.
+Our first step is to list out all the books we know about.
 
 Let's write a new feature test describing what we want to achieve:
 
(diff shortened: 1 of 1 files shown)


content/introduction/getting-started.md
```diff
@@ -112,33 +112,33 @@ And... bask in the glory of your first Hanami project at
...
 They live under `apps/`.
 
-By default, we have a `web` app, which can be thought of as the normal, user-facing web interface.
+By default, we have a `web` app, which can be thought of as the standard, user-facing web interface.
 This is the most popular, so you'll probably want to keep it in your future Hanami projects.
-But, just so you know, there's nothing special or different about this app, it's just so common that Hanami generates it for us.
+However, there's nothing unique about this app, it's just so common that Hanami generates it for us.
 
-Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `api`, or an analytics `dashboard`.
+Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `API`, or an analytics `dashboard`.
```
cllns:
```suggestion
Later (in a real project), we would add other apps, such as an `admin` panel, a JSON `api`, or an analytics `dashboard`.
```
I agree API should be capitalized, but this context is the name of the app, which _has_(?) to be lowercase.

content/introduction/getting-started.md
```diff
@@ -112,33 +112,33 @@ And... bask in the glory of your first Hanami project at
...
 We could also break our `web` app into smaller.
```
cllns:
```suggestion
We could also break our `web` app into smaller parts, extracting isolated pieces of functionality.
```
Just noticed this sentence was incomplete :)

content/introduction/getting-started.md
```diff
@@ -112,33 +112,33 @@ And... bask in the glory of your first Hanami project at
...
 Hanami fully supports that, too!
 
 Different `apps` represent __delivery mechanisms__.
 That means they're different ways of interacting with the core of your project, or the "business logic".
 
-Hanami doesn't want us to [repeat ourselves](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), so that "business logic" is shared.
+Hanami doesn't want us to [repeat ourselves](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) so that "business logic" is shared.
```
cllns:
```suggestion
Hanami doesn't want us to [repeat ourselves](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), so that "business logic" is shared.
```
I think the comma helps here a lot, and changes the meaing of the sentence. Perhaps we could find another way to re-word this so that it's more clear and we can avoid a single comma changing the meaning. What do you think?

content/introduction/getting-started.md
```diff
@@ -226,8 +226,8 @@ end
 ```
 
 This is an empty action that doesn't do anything special.
-Each action in Hanami is defined by [its own class](https://en.wikipedia.org/wiki/Single_responsibility_principle), which makes it simple to test and work on.
-And, each action has a corresponding view, which is also defined by its own class.
+Each action in Hanami is defined by [its class](https://en.wikipedia.org/wiki/Single_responsibility_principle), which makes it simple to test.
```
cllns:
```suggestion
Each action in Hanami is defined by [a single class](https://en.wikipedia.org/wiki/Single_responsibility_principle), which makes it simple to test.
```
I think this suggestion gets to the point we're trying to make here. (That it's different than Rails where you can have 7+ actions in a class)

content/introduction/getting-started.md
```diff
@@ -326,7 +325,7 @@ This does a lot for us:
 - Creating a view at `apps/web/views/books/index.rb` (and a spec for it),
 - Creating a template at `apps/web/templates/books/index.html.erb`.
 
-_(If you're confused by 'action' vs. 'controller', here's a tip: Hanami only has `action` classes, so a controller is just a module to group several related actions together.)_
+_(If you're confused by 'action' vs. 'controller,' here's a tip: Hanami only has `action` classes, so a controller is just a module to group several related actions together.)_
```
cllns:
```suggestion
_(If you're confused by 'action' vs. 'controller': Hanami only has `action` classes, so a controller is just a module to group several related actions together.)_
```

content/introduction/getting-started.md
```diff
@@ -412,7 +411,7 @@ There are two types of objects that we'll use for this:
 * an **entity** is a domain object (a `Book`) that is uniquely identified by its identity,
 * a **repository** is what we use to persist, retrieve, and delete data for an entity, in the persistence layer.
 
-Entities are totally unaware of the database.
+Entities are unaware of the database.
```
cllns: Hm, I think this is a pretty big distinction so I like the adverb for emphasis. Maybe `entirely` instead of `totally`, what do you think?

content/introduction/getting-started.md
```diff
@@ -473,10 +472,10 @@ $ HANAMI_ENV=test bundle exec hanami db prepare
 
 ### Working With Entities
 
-An entity is something really close to a plain Ruby object.
+An entity is something really close to an everyday Ruby object.
```
cllns:
```suggestion
An entity is something really close to a plain Ruby object.
```
I think the word `plain` is better here, since "Plain old Ruby object" is [a common term](https://codesthq.com/hi-im-poro/), and often abbreviated as "PORO" (though I don't think we shoud introduce that acronym here).

And it comes from Java, where it's also [a common term](https://en.wikipedia.org/wiki/Plain_old_Java_object)

content/introduction/getting-started.md
```diff
@@ -501,7 +500,7 @@ end
 ### Using Repositories
 
 Now we are ready to play around with our repository.
-We can use Hanami's `console` command to launch IRb with our application pre-loaded, so we can use our objects:
+We can use Hanami's `console` command to launch irb with our application pre-loaded, so we can use our objects:
```
cllns:
```suggestion
We can use Hanami's `console` command to launch `irb` with our application pre-loaded, so we can use our objects:
```

content/introduction/getting-started.md
jacobherrington: Why do you want to avoid using PORO? That would have clarified it for me a bit.

jacobherrington:
@cllns your comments have been addressed. Let me know what you think 👍

cllns:
Thank you @jacobherrington, awesome job.

PR #18, opened 2018-12-18
Add example of generating nested actions

aeden:
I was missing an example that shows how to generate a nested action for nested resourceful routes. This PR adds an example in the generator guide.

### DIFF
--- a/content/command-line/generators.md
+++ b/content/command-line/generators.md
@@ -30,13 +30,19 @@ The first argument, `web`, is the name of the target application in a Hanami pro
 
 The argument `books#show` is the name of the controller and the action separated by the number sign (`#`).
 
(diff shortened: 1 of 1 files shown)


content/command-line/generators.md
```diff
@@ -30,13 +30,19 @@ The first argument, `web`, is the name of the target application in a Hanami pro
 
 The argument `books#show` is the name of the controller and the action separated by the number sign (`#`).
 
+For nested actions, use a slash (`/`) to separate the modules. Nested actions work in conjunction with nested resourceful routes. For example:
+
+```shell
+$ bundle exec hanami generate action web books/editions#show
+```
```
davydovanton: WDYT for adding example of generated action? I think it will be helpful for showing what user will get in this case

aeden: Done.

cllns:
Thanks!

PR #20, opened 2019-01-06
Expect not to create a book if invalid

djuber:
When `params` are invalid, we don't expect to call the `AddBook` interactor. Since the controller is performing validation - the interactor is not used.

Change spec to match this, assert that interactor does not receive call.

### DIFF
--- a/content/architecture/interactors.md
+++ b/content/architecture/interactors.md
@@ -554,7 +554,7 @@ RSpec.describe Web::Controllers::Books::Create do
   context 'with invalid params' do
     let(:params) { Hash[book: {}] }
 
-    it 'calls interactor' do
+    it 'does not call interactor' do
       expect(interactor).to_not receive(:call)
       response = action.call(params)
     end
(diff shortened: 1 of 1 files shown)


content/architecture/interactors.md
```diff
@@ -555,7 +555,7 @@ RSpec.describe Web::Controllers::Books::Create do
     let(:params) { Hash[book: {}] }
 
     it 'calls interactor' do
```
cllns:
```suggestion
    it 'does not call interactor' do
```

cllns:
Thanks for this! I used GitHub's suggestions feature to propose fixing the previous line too. Can you approve that change? Once that's in, looks good to me :+1:

djuber:
> Thanks for this! I used GitHub's suggestions feature to propose fixing the previous line too. Can you approve that change? Once that's in, looks good to me +1

Sorry for the delay - merged in your suggestion, checking #22 to make sure this is still relevant.

cllns:
Thanks @djuber! 🌸

PR #22, opened 2019-01-21
Use described_class in create book controller spec

sharipov-ru:
In the Getting Started Guide -> [Implementing Create Action ](https://github.com/hanami/guides/blame/master/content/introduction/getting-started.md#L787) section when `create_spec.rb` file is originally created, `let(:action)` is defined as  `described_class.new`. But when we proceed to Architecture -> Interactors guide, then we notice that `let(:action)` is set to the full name of the controller. This PR changes it to `described_class.new` so that it feels less confusing when you read Getting Started Guide and then Architecture -> Interactors guide and compare changes.

### DIFF
--- a/content/architecture/interactors.md
+++ b/content/architecture/interactors.md
@@ -554,8 +554,8 @@ RSpec.describe Web::Controllers::Books::Create do
   context 'with invalid params' do
     let(:params) { Hash[book: {}] }
 
-    it 'calls interactor' do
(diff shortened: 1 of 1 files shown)


sharipov-ru:
Just noticed that there is already existing PR for the interactor spec fix, so reverted corresponding updates, only `let(:action)` change is proposed now.

davydovanton:
thanks for PR!

Issue #9, opened 2018-10-27
Text a little hard to read in certain classes with background colors


ecnal:
The new Guides are beautiful.  I do have a problem reading text in notice and convention classes.  It's more readable if the font-weight is higher and/or the background color is darker.

jodosha:
@ecnal Could you please send us a screenshot attached in this issue? Thanks.

ecnal:
![screenshot from 2018-10-30 11-32-47](https://user-images.githubusercontent.com/1310070/47730033-adae6680-dc37-11e8-8f1d-54af6c45bd88.png)
![screenshot from 2018-10-30 11-32-24](https://user-images.githubusercontent.com/1310070/47730056-bc951900-dc37-11e8-8238-4cd95218c3b9.png)
![screenshot from 2018-10-30 11-31-24](https://user-images.githubusercontent.com/1310070/47730067-c1f26380-dc37-11e8-9ef2-322e544945f5.png)

whysthatso:
I have the same issue and will provide a pr that might make it a bit more legible.

jodosha:
@ecnal Thanks for closing this. I hope the text is more readable for you now.

jodosha:
Fixed by https://github.com/hanami/guides/pull/30

PR #1, opened 2018-10-16
Use different colour for code blocks and for notify block text

davydovanton:
Just change `code` colour and text colour for all notifications blocks

## Screenshots
screenshot 2018-10-16 13 03 18
screenshot 2018-10-16 13 14 40
screenshot 2018-10-16 13 14 45
screenshot 2018-10-16 13 14 54

PR #30, opened 2019-02-23
improve readability of flash-like block elements

whysthatso:
i hope this does not shoot through some global design styles, however, i do think it makes the site much more legible. colors are rather randomly darkened.

alternative: darken the font color for these elements.

see #9 

current:

screenshot from 2019-02-23 20-16-48

changed:

screenshot from 2019-02-23 20-19-16

### DIFF
--- a/themes/hanami/static/assets/css/theme.css
+++ b/themes/hanami/static/assets/css/theme.css
@@ -242,20 +242,20 @@ p.warning {
 
 p.notice {
   color: #FFF;
-  border-color: #35DDFF;
-  background-color: #35DDFF;
+  border-color: #27A8C2;
+  background-color: #27A8C2;
 }
 
 p.convention {
   color: #fff;
-  border-color: #46E895;
-  background-color: #46E895;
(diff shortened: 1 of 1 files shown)


kaikuchn:
Hi Andreas,

thank you for taking the time to improve our guides! I agree with you that the current color scheme has some rather unfortunate contrasts which make parts of the guide illegible - especially on some uncalibrated screens. We had raised this issue internally before, but somehow never got around to solving it (see #1 ).
The color palette was taken from here https://color.adobe.com/de/My-Color-Theme-color-theme-11269714/ and while I agree that we need to do something about the readability we should not break the cohesiveness of the colors - i.e., we should stay in one palette. Maybe change the whole palette? Or find a way to stick to the current palette, and still improve legibility.

kaikuchn:
Looks good on my iPhone, will check it out later in other screens.

whysthatso:
i got a bit lost with all the patch branches. reverting changes and merging in new stuff is not so straight forward. but i think this is ready now.

PR #34, opened 2019-03-04
Add information about :logger mailer mode

katafrakt:
Adds information about `:logger` mode for mailer. This is very useful for development and I was quite upset it's not available. Then I wrote my own mode and then I discovered it was already available, but undocumented.

By the way, I think this should be default mode in development environment, so I'm thinking about adding it as default in `unstable` branch of `hanami/hanami` - WDYT?

### DIFF
--- a/content/mailers/delivery.md
+++ b/content/mailers/delivery.md
@@ -49,6 +49,7 @@ It accepts a symbol that is translated into a delivery strategy:
   * Sendmail (`:sendmail`)
   * SMTP (`:smtp`, for local SMTP installations)
   * SMTP Connection (`:smtp_connection`, via `Net::SMTP` - for remote SMTP installations)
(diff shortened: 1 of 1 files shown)


kaikuchn:
Hey, that is neat!

I agree that this may be a useful default for development. Maybe the [template](https://github.com/hanami/hanami/blob/8d0ae2e5ed2bc06964ed0b5bdc95c2908b814fde/lib/hanami/cli/commands/new/config/environment.erb) for the `environment.rb` could be adapted accordingly?

katafrakt:
@kaikuchn Yes, that was my thought - to change template to

```
  environment :development do
    # See: https://guides.hanamirb.org/projects/logging
    logger level: :debug
    
    mailer do
      delivery :logger
    end
  end
```

I'll prepare the PR.

PR #36, opened 2019-04-04
Update getting-started.md

poudelmadhav:
homes is renamed to home


davydovanton:
hey, I have a question. Are you sure that we need to use `home` for controller name instead `homes`? @hanami/core WDYT?

poudelmadhav:
As you have previously created the controller name `home`, but at this line you have used `homes`. So I have suggesting you.

davydovanton:
Got it! Good catch 👏

poudelmadhav:
Welcome

Issue #17, opened 2018-12-15
Add guidance on how to test when cookies and sessions are enabled


aeden:
While reading through the Actions: Sessions section, I miss guidance on how to properly test this. I did find an article that describes using `rack.session` as a param, but it'd be great to have an official way to test in controllers when optional functionality like sessions and cookies are enabled.

kaikuchn:
Hey Anthony,

thanks for raising this issue. If I recall, this or similar questions have been asked in the Chat. So I think you are right, this would be a good thing to add.

Would you like to work on a PR for this?

aeden:
Unfortunately I do not have time to work on a PR for this now.

kaikuchn:
Ok, no worries. I'll put it into my backlog. If you happen to start working on this, just ping use here. If I or somebody else found the time to work on it, we'll do the same.

davydovanton:
@kaikuchn I'll create a PR because I have a lot of experience with testing sessions

davydovanton:
@kaikuchn @aeden could you check my PR?

PR #38, opened 2019-04-04
Add information about session in action tests

davydovanton:
Related to https://github.com/hanami/guides/issues/17

### DIFF
--- a/content/actions/testing.md
+++ b/content/actions/testing.md
@@ -306,3 +306,59 @@ end
 <p class="notice">
 Please avoid <em>test doubles</em> when writing full integration tests, as we want to verify that the whole stack is behaving as expected.
 </p>
+
+## Session Testing
+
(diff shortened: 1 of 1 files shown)


content/actions/testing.md
```diff
@@ -306,3 +306,59 @@ end
 <p class="notice">
 Please avoid <em>test doubles</em> when writing full integration tests, as we want to verify that the whole stack is behaving as expected.
 </p>
+
+## Session Testing
+
+If action works with session and you need to test this action you can put session to `params` block. For example you have a action with current account:
```
siaw23: I think it'd be better put this way:

"If you need to test an action that works with a session, you can place the session in a `params` block. For example, say you have an action with a `current_account`:"

davydovanton:
@siaw23 hey man, thanks for review and comments. I fix it 💯

davydovanton:
Thanks for review!

PR #39, opened 2019-04-08
Fix some wording

siaw23:
Just a few things I found worth fixing.

### DIFF
--- a/content/introduction/getting-started.md
+++ b/content/introduction/getting-started.md
@@ -205,7 +205,7 @@ root to: 'home#index'
 
 We pointed our app's root URL to the `index` action of the `home` controller (see the [routing guide](/routing/overview) for more information).
 
-If we run our tests, we'll get an error that the endpoint cannot be found.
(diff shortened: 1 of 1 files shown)


content/introduction/getting-started.md
```diff
@@ -483,7 +483,7 @@ class Book < Hanami::Entity
 end
 ```
 
-This class will generate getters and setters for each attribute which we pass to initialize params.
+This class will generate getters and setters for each attribute which we pass to initialize `params`.
```
kaikuchn:
```suggestion
This class will generate getters and setters for each attribute we pass to initialize `params`.
```

siaw23: I don't understand this one. What do you mean? "which" has not been removed in this PR.

kaikuchn: Yeah, it's the change I suggest. I think it should be removed.

kaikuchn:
Thanks again for your contribution!

PR #56, opened 2019-12-17
Fix code sample for extending existing CLI commands

katafrakt:
Not sure if it's something new (because of renaming `hanami-cli` to `dry-cli`?) - as I think it worked before. But now for freshly generated Hanami project (version 1.3.3) I needed to extend via `Hanami::CLI::Commands`.


adam12:
Thanks for the patch!

PR #68, opened 2020-07-23
Change book title in `add_book_spec.rb`

jtanium:
Small thing: the `add_book_spec.rb` can give a false positive because the link text to create a book is "New book" in `apps/web/templates/books/index.html.erb`, so even if the add failed and `books/index.html.erb` is rendered, this test would pass. I modified the title to differentiate it from the link text.


cllns:
Thank you, great find 👍  🌸

PR #69, opened 2020-07-24
Update Gemfile snippet in preprocessors.md

Bajena:
Hey, I've tried building a simple app with SCSS assets compilation, but when I added the `sass` gem as suggested in the docs it was crashing with following error:
image

it seems that the `sass` gem is deprecated and the authors suggest using `sassc`. After this change the assets started compiling normally.


cllns:
:+1: Thank you!

PR #70, opened 2020-08-12
Improve Session Testing with example of session OUT values

francois:
After reading the Hanami Guide from A to Z, I did not see any example describing how to assert values coming *from* the action. The only example showed how to pass values *into* the action.

Thanks to @kaikuchn for the [tip that unstuck me](https://gitter.im/hanami/chat?at=5f33bc5d79659c5fbfeb2021)!


francois:
Direct link to changes: https://deploy-preview-70--hanami-guides.netlify.app/actions/testing/#session-testing

PR #76, opened 2021-02-16
Update the html5 guide for link_to

### LINKED ISSUE #51: link_to helpers and concat
The guide indicates that `link_to` helpers do not need to be concated:

    html.div do
      link_to 'Users', routes.users_path, class: 'btn'
      hr
      link_to 'Books', routes.books_path, class: 'btn'
    end

However, in practice, only the last link renders. Upon some digging, apparently they need to be concated:

https://github.com/hanami/helpers/blob/9acd2bf17bc2e5579cdd91f8d0f88129367cb420/spec/support/fixtures.rb#L715

I'm going to assume that the code is correct and the guide needs updating?

adam12:
The `link_to` helper will not append itself to the output buffer. A node
or pseudo node (ie. text) is required to have it rendered.

Closes #51


content/helpers/html5.md
```diff
@@ -131,27 +131,50 @@ html.tag(:custom, 'Foo', id: 'next') # => <custom id="next">Foo</custom>
 html.empty_tag(:xr, id: 'next')      # => <xr id="next">
 ```
 
-## Other helpers
+## Link helpers
 
-Hanami html helpers also support other assembled helpers. For example `link_to` helper:
+The `link_to` helper, while appearing similar to tags generated by `html`, have different semantics and will
```
cllns, line 136:
```suggestion
The `link_to` helper, while appearing similar to tags generated by `html`, has different semantics and will
```

cllns (approved):
Approved with one small fix!

PR #78, opened 2021-04-08
Update sessions.md

sebastjan-hribar:
This update is about short instruction for sharing session values.

### DIFF
--- a/content/actions/sessions.md
+++ b/content/actions/sessions.md
@@ -64,3 +64,47 @@ end
 ### Sharing session values between applications
 
 To share session values, defined in one application, we must provide the same session secret to all the applications where we need those values.
+
+```ruby
+# Define ENV variables for development environment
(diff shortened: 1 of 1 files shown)


content/actions/sessions.md
```diff
@@ -61,3 +61,6 @@ module Web
   end
 end
 ```
+### Sharing session values between applications
+
+To share session values, defined in one application, we must provide the same session secret to all the applications where we need those values.
```
adam12, line 66: Would it be worth duplicating the example code block above to this section, and then displaying two application classes with the same session key set for both?

sebastjan-hribar: Thank you for the suggestion, I've made the changes.

adam12:
Sorry for the delay on this. Are you able to rebase @sebastjan-hribar?

adam12:
Closing in favour of #104

PR #79, opened 2021-04-08
Update forms.md

sebastjan-hribar:
Instructions on how to generate and pass CSRF token with UJS,


cllns:
Thanks @sebastjan-hribar 🌸 

I think putting a note here about CSRF is a good idea. The two solutions are to [disable CSRF protection for the action](https://github.com/hanami/hanami/blob/master/lib/hanami/action/csrf_protection.rb#L116) or to pass the CSRF token like you do here. I think the former is generally better, but it may not be possible if you're using the action for traditional form submission. 

However, we should _not_ assume users are using jQuery or any other specific JS libraries. Maybe this could just be a note saying what should be done (maybe with a link to stackoverflow to help users?). That is, just say where the CSRF token is located `<meta name="csrf-token" content="...">` and where it should be inserted into a request (`_csrf_token`). Then the users can put that value into the request themselves.

Can you change this PR to outline the two options?

sebastjan-hribar:
Hi, @cllns,
I'm sorry for not replying sooner. I'm afraid I don't understand the context of your comment. Initially I've disabled the CSRF protection to get our AJAX calls to work. But I would never opt to disable it for no special reason.

Also, I might phrased my text above a bit awkward, but I wouldn't assume anyone is using UJS. I merely intended the people to know how to go about it **if** they use UJS. I'm reluctant to say it is possible to disable it when this might expose some vulnerabilities, but then again, I'm not a security expert. If you still believe the 2 options of handling with CSRF need to be present, I can make the change.
Best, Sebastjan

adam12:
I think this might be more appropriate as a [Hanami Snippet](https://github.com/hanami/snippets). Please submit a PR there if you feel the same way.

PR #81, opened 2021-04-16
Use 'now' for year, rather than hard-coded

cllns:
I don't know where / if the footer file is used, but these were the two places where `2018` appeared. 

I know that, for copyright purposes, `2018` is fine but I think it looks better to have the most recent year in the footer.

Happy to change it to `2018`-`{current year}`, if you prefer (or earlier even)

(If you're reading this and new to Go, like I am, [`2006` is a special formatting value that means `%Y` in strftime](https://golang.org/pkg/time/#pkg-constants))


jodosha (approved):
@cllns Good one 👍

PR #82, opened 2021-04-16
Improve design

### LINKED ISSUE #84: Left-aligned content looks better then justified
image

cllns:
Before:
image

After:
image

These screenshots are 1920px wide. 
Displayed on a 27 inch monitor:
- Before: the main column is ~28cm (11 inches) wide. ~125 characters per line
- After: the main column is ~20cm (8 inches) wide. ~100-110 characters per line

This is still much higher than the recommended maximum number of characters per line (~75-80) but I think it looks fine. Happy to make the column even narrower. Right now we get ~20% more content per screen with this design, and that'll be reduced if the content is narrower.

You can walk through the commit messages to see the details of what was changed. But some highlights:
- Narrow the width of the content to 2/3rds of the center column rather than 100%
- Change font from Open Sans to the OS native stack. This is a very common practice now.
- Lower size from 1.2rem (19.2px) to 1rem (16px).
- Make the notice/warning/convention color less intense by adding transparency.
- Make the code-highlighting (yellow background) less intense, and reduce padding too.

A couple things:

- I wrapped the main content in another div to keep the fixed sidebar all the way on the right side where it belongs. Another way to do this is to create empty divs that only have width at xl (and larger) width. The approach I did here of having an internal column system that's centered and 8/12 wide seems to work fine though.

- ~For some reason my local `hugo` doesn't render any of the embedded HTML that we use in markdown for  `<p>` tags with classes `notice`, `warning`, or `convention`. I'm not sure why this happens, but I guess it might be related to having a newer hugo version than the build server uses? I did update those colors, by changing them in the web inspector on the production site, to make sure the colors look good. I can make sure those render once this is merged.~ UPDATE: This is all still true but it generated fine with the preview from Netlify, so I think we'lll be all set.

Happy to revisit any of these, or look at any other parts of the design.

(Edit: Closes #84)

### DIFF
--- a/themes/hanami/layouts/_default/single.html
+++ b/themes/hanami/layouts/_default/single.html
@@ -141,52 +141,54 @@
         </ul>
       </div>
       <main class="col-12 col-md-9 col-xl-8 py-md-3 pl-md-5 ct-content">
-        <div class="ct-page-title">
-          <h1 class="ct-title" id="content">{{ .Section | humanize | title }}: {{ .Title }}</h1>
-          <div class="avatar-group mt-3">
-          </div>
-        </div>
-        <hr>
-        {{ .Content }}
-
-        {{- $counter := 0 -}}
-        {{- $index := -1 -}}
-        {{- $pages := slice -}}
-        {{- range .Site.Params.topics -}}
-          {{- range sort (where $.Site.RegularPages "Section" .) ".Params.order" -}}
-            {{- $counter = add $counter 1 -}}
-            {{- if eq .URL $.Page.URL -}}
-            {{- $index = (sub $counter 1) -}}
-            {{- end -}}
-
-            {{- $pages = $pages | append . -}}
-          {{- end -}}
-        {{- end -}}
-        {{- $previousPage := index $pages (sub $index 1) -}}
-        {{- $nextPage := index $pages (add $index 1) -}}
-
-        <hr>
-        <div class="container">
-          <div class="row">
-            <div class="col">
-              {{ with $previousPage }}
-              <a href="{{ .URL }}">
-                <button class="btn btn-icon btn-3 btn-outline-primary" type="button">
-                  <span class="btn-inner--icon"><i class="ni ni-bold-left"></i></span>
-                  <span class="btn-inner--text">{{ .Section | humanize }}: {{ .Title }}</span>
-                </button>
-              </a>
-              {{ end }}
+        <div class="col-xl-8 mx-xl-auto">
+          <div class="ct-page-title">
+            <h1 class="ct-title" id="content">{{ .Section | humanize | title }}: {{ .Title }}</h1>
+            <div class="avatar-group mt-3">
             </div>
-            <div class="col text-align-right">
-              {{ with $nextPage }}
-              <a href="{{ .URL }}">
-                <button class="btn btn-icon btn-3 btn-outline-primary" type="button">
-                  <span class="btn-inner--text">{{ .Section | humanize }}: {{ .Title }}</span>
-                  <span class="btn-inner--icon"><i class="ni ni-bold-right"></i></span>
-                </button>
-              </a>
-              {{ end }}
+          </div>
+          <hr>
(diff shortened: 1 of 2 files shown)


cllns:
For some reason [the first build above failed, with an error saying Netlify couldn't get Go v1.10](https://app.netlify.com/sites/hanami-guides/deploys/6079fa7934fc060007292c44)

Just merged a tiny commit from master to re-trigger the build and it worked fine.

Can preview these changes here: https://deploy-preview-82--hanami-guides.netlify.app/introduction/getting-started/

adam12:
How do you feel about just limiting the content inside of the entire column? And keeping the header and any footer navigation full width? The narrow column ends up with some edge cases such as the one below.

image

If I'm being honest, I'd probably prefer the content column to be left aligned. When centered it appears to be disconnected and not attached to anything.

cllns:
Thanks for looking at it :)

### Footer
Good call with the footer! I didn't see that wrapping issue. It actually already happens with the existing site, at page width of ~875 pixels and fewer.

This PR did have the issue, at around ~930 pixels and fewer. I just added a commit that pulls the footer navigation outside the narrow column and the issue still happens, but at ~680 pixels and less now. 

I also added a commit that adds some vertical padding to them, which fixes the issue well enough IMO (since we _do_ want the back to be on the left and the next to be on the right).

### Alignment / Header
As far as the header goes, it looks a little strange with the header on the left and the content in the middle. Especially because of the purple left border (I added 20% transparency to it regardless).


I just tried it out making the whole content column left aligned doesn't look as bad as I thought it would. I'd be OK with that, but I still prefer the balance of having it in the center. Let's see what others say, I'm cool with either :)

baweaver:
Looking at the design a bit myself, and would agree with a lot of the current observations with a few additions:

* Consider removing the highlight on inline code entirely, replace with a mid-dark red
* Consider darkening the header as there's not much contrast there (around `611cb4`-ish)
* Same with the buttons
* Consider making headers clickable, and a darker version of the link color for some more visual separation

On the list of relatively minor things it may be worthwhile to use variables instead of a header comment for color schemes to make it easier to change later.

cllns:
@baweaver Thanks for all that feedback! Unforunately I can't seem to figure out how to change the code block styles. They're defined in a `docs.min.css`, which I don't want to edit directly. It seems like the source for it is coming from some [argon](https://demos.creative-tim.com/argon-design-system/) file, b

Similarly, changing the font-family or highlighting (without resorting to `!important` or piecemeal overriding `code[class*="language-"]` rules) didn't seem obvious to me.

I do think the header/buttons looks better with a darker color, but it's also a non-negligible aesthetic change, so I'd like @jodosha to make the call on that.

I really do appreciate your feedback and think it would improve the design, I just can't figure out how to do it quickly. I can dive in more to if if others request it, but I figure the existing changes here get us some lower-hanging fruit.

parndt:
I noticed that the main page (https://guides.hanamirb.org/) uses a `max-width` of `1040px` when viewing on resolutions greater than `1200px`, with automatic horizontal margins. What do you think about making the documentation pages match this design?  This would look like:

image

jodosha (commented):
@cllns I'm fine with the changes, please work with the rest of the team to manually verify that the design works. 👏 🙂

cllns:
I don't really like the white gutters on the side. It feels very constrained. Not sure which element you were thinking of but when I put `container` on `body` (to check, now what we'd want to do, it looks and feels too narrow). Here's what it looks like on my 4k monitor: image

I could be into doing a max-width, but I think the top header should be full-width at least. Still, I think having a design that takes advantage of the full screen width is preferable

parndt:
@cllns all good, thanks for the consideration; I'm not interested in blocking what you have, it's a readability improvement. 👍

At some point (not today) it could be nice to bring hanamirb.org, guides.hanamirb.org, and the documentation pages inside the guides site to all use a similar design which would make them feel more connected.

cllns:
Closing as stale, many other PR's have improved the design in various ways

PR #85, opened 2021-05-17
Version prefix in URL and single page titles

jodosha:
### Prefixed URLs

Contents are prefixed in URLs with `vMAJOR.MINOR` Hanami version. It was `/actions/overview/`, it's now `/v1.3/actions/overview/`.

This change will make possible to document multiple Hanami versions.

### Old URLs

Old URLs (non prefixed) are **temporary** redirected to the prefixed counterpart. E.g. `/actions/overview/` redirects to `/v1.3/actions/overview/`.

This redirect should be removed once we will release 2.0 guides.

### Page titles

Now page titles have the version explicitly set: e.g. `v1.3: Actions`.

### Get Started button

The "Get Started" button on the home page now points to the latest version of Hanami (e.g. `/v1.3/introduction/getting-started/`.

In `config.toml` there is `hanamiversion`, which right now is `1.3`.

### Sidebar sections

Sidebar sections are now ordered according to the file system metadata, and no longer on `topics` setting in `config.toml`. This allows to have multiple versions with different sections.

Each section has its own Hugo file to indicate the section title and order.

Example:

```bash
⚡ cat content/v1.3/entities/_index.md
---
title: Entities
order: 90
---
```

Entities section will take position 9 in the sidebar. Order has steps with increments of 10, to easily insert new sections without the need of manually change the order for all the sections.


swilgosz:
This is amazing! @jodosha but what about redirecting existing URLs? We don't want to loose all the existing reference links, do we?

swilgosz:
@jodosha After adding v2.0 folder with Documents, and changing the .toml configuration to point to v2.0, the sidebar shows duplicated files list.



image

jodosha:
@swilgosz Fixed both the things that you pointed out.

PR #86, opened 2021-05-22
Fix broken image

baweaver:
Welcome page image is broken, quick fix.


adam12:
Doesn't look like this worked. I'd probably just prepend `/v1.3` to the original path. WDYT?

PR #88, opened 2021-06-04
Multilevel sections

jodosha:
## Feature

Allow three levels of nesting in sidebar.

Tried it with Command Line section and split it into many articles.


swilgosz:
@jodosha No, moving content in different PR.

PR #90, opened 2021-06-09
Add Guides 2.0

swilgosz:
# Overview

Guides for Hanami 2.0

- [ ] Getting Started
- [x] Architecture
    - [x] Overview
- [ ] Router 
    - [x]  Overview
    - [x] Basic Usage
    - [x] Advanced Usage
    - [ ] Testing
- [ ] Actions
- [ ] Views
- [ ] Persistance
- [ ] Validations
- [ ] Slices
- [ ] Assets
- [ ] Utils
- [ ] Helpers
- [ ] Mailers
- [ ] Command Line
- [ ] Upgrade Notes


content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+Hanami.application.routes do
+  slice :main, at: "/" do
+    root to: "home.show"
+
+    get '/hello', to: ->(env) { [200, {}, ['Hello from Hanami']] }
+  end
+end
+```
+
+Then start the server with `puma config.ru` and visit [http://localhost:3000/hello](http://localhost:2300/hello).
```
cllns, line 32:
```suggestion
Then start the server with `bundle exec rackup` and visit [http://localhost:3000/hello](http://localhost:2300/hello).
```

I'm not sure but I think this will do the same thing when you have puma configured, and run whatever other server you have configured if you don't.

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+You should see `Hello from Hanami!` in your browser.
+
+<p><img src="/v2.0/router/hello-from-hanami.png" alt="Hanami welcome page" class="img-responsive"></p>
```
cllns, line 35: I don't think this screenshot is particularly worth including here. It's just rendering plain-text so there's no visual elements. The browser chrome will inevitably be different throughout the guides, with different contributors using their own so minimizing then number of times we show browser screenshots is worth doing, I think :)

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+Let's explain what we just did.
+We created a **route**; an application can have many routes. By doing this, we told our app, to handle HTTP `get` request under the path: `/hello`, and handle it by calling a `Proc` object, that returns the status: `200`, empty headers, and the `'Hello from Hanami'` body. This response is given back to the browser for rendering.
```
cllns, line 38:
```suggestion
We created a **route**; an application can have many routes. By doing this, we told our app to handle HTTP `get` request under the path `/hello`, and handle it by calling a lambda that returns the status code `200` (a success), with empty headers, and the `'Hello from Hanami'` body. This response is given back to the browser for rendering.
```
A lambda is an instance of `Proc`, but they behave differently. Since Hanami::Router is agnostic about which it received (I assume 😇 ), I think saying that it's a `Proc` might make people think it wants non-lambda Procs

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+Each route starts with an [HTTP verb](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) declaration, `get` in our case.
+Then we specify a relative URI (`/hello` for us) and the object that is responsible to respond to incoming requests.
```
cllns, line 41:
```suggestion
Then we specify a relative URI (`/hello` for us) and the object that is responsible for responding to requests.
```

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+**That is pretty neat!**
```
cllns, line 43: It is neat! (And I like this style of interjecting some enthusiasm 🙂

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+The only requirement for our routes to be handled properly, is to pass an object as a handler, that responds to: `call` method, accepting one argument, and returning the expected response object.
```
cllns, line 45:
```suggestion
The only requirement for our routes to be handled properly is for the object that handles the request to implement a `call` method that accepts one argument (the Rack `env`), and that it returns the expected response format.
```

### Also
I'm not sure if we should get into the Rack expected response format here (the Array with size 3 with status code as Int, Headers as `Hash` and Body as Array (of Strings)).

The reason I'm not sure is that full-stack Hanami apps will call instances of `Hanami::Action` most of the time. They do this too, but indirectly.

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+Here are two different implementations of the same thing, check it out!
```
cllns, line 47: I'm not sure if this example is worth including. I get the point, but we don't want to encourage people using class methods like this :D 

Maybe instead, this can be an example of integrating with a `Hanami::Action`? And mixing a lambda with a Hanami action in the same set of routes?

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,73 @@
...
+get '/hello', to: ->(env) { [200, {}, ['Hello from Hanami!']] }
+```
+
+So the whole file would look like this:
+
+```ruby
+# /config/routes.rb
+
+Hanami.application.routes do
+  slice :main, at: "/" do
```
cllns, line 24: Should we explain how slice routing works here too, at least briefly? There will be more guides about slices, of course, but this will be a new concept to almost everyone so briefly mentioning it's something new to learn could be useful (though the routing portion here is almost trivial and obvious).

cllns (commented):
🎉  Excited we're getting started with these 2.0 guides. Great work :)

content/v2.0/router/overview.md
swilgosz: I believe this will be changed to sth like `hanami start` relatively soon. The CLI is not yet done, but I don't expect these guides to be finished earlier. @cllns what do you think?

content/v2.0/router/overview.md
swilgosz: I see it as a placeholder for now. The 1.3 Hanami new app had a pre-styled generated home page, which was actually worth being presented on the screenshot. I'd rather suggest changing it later to the default, instead of completely removing it.

content/v2.0/router/overview.md
swilgosz: Lol. Didn't know! :D Thanks! By writing these docs I'll learn more advanced Ruby's tips, than in a year of regular development ;).

content/v2.0/router/overview.md
swilgosz: Hmm... I'll process all docs through Grammarly first, before giving them to review ;).

content/v2.0/router/overview.md
swilgosz: My idea was to allow a user to type sth from scratch, without initializing Magic objects, that works out of the box. This is a very minimal setup, very easy to understand and to present the core of the behavior.

I'm moving towards actions in the `basic_usage` section. Soon I should push the remaining two files (`basic_usage` and `advanced_usage`) for the router, then we'll have a better overview, of how to reorder the content.

content/v2.0/router/overview.md
swilgosz: Yeah, I actually agree. But can we use Hanami::Action without rendering the view? - Just passing the returned value? If I remember, it's by setting the `self.body` ?

content/v2.0/router/overview.md
swilgosz: This is also a topic I briefly mentioned in the `basic_usage` - Let's discuss if it's better to put it here when I'll push remaining docs :). I've not a strong opinion about it yet.

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,221 @@
...
+  end
+  
+  slice :admin, at: "/" do
+    root to: "dashboard.show" # slices/admin/lib/actions/dashboard/show.rb
+    get '/users', to: 'users.index' # slices/admin/lib/actions/users/index.rb
+  end
+end
+```
+
+Read more about [slices](/v2.0/slices/overview)!
```
swilgosz, line 111: @timriley is it more to this to include here? Or is it fine to put the reset in the `slices` section` ?

content/v2.0/router/overview.md
cllns: Yea that makes sense. Can we come up with some kind of tag to make sure we go through and update these? `<!-- FIXME: Change to `bundle exec hanami server` when it's ready -->`
Though in this case, we know exactly what the command will be, so maybe we should just write it in here, even though it doesn't work yet.

content/v2.0/router/overview.md
cllns: Agree it's useful for the Getting Started guide, but this is just a simple lambda endpoint. 

Even with [the pre-styled default page that we have in the existing Getting Started guide](https://guides.hanamirb.org/v1.3/introduction/welcome-page.png) implemented, this simple lambda endpoint will never render a view or a template, it's just returning text.

content/v2.0/router/overview.md
cllns: Yup! `response.body =` though, instead of `self.body =` (since the action is frozen)

content/v2.0/router/advanced-usage.md
```diff
@@ -2,3 +2,52 @@
 title: "Router: Advanced Usage"
 order: 30
 ---
+
+
+[Refer](https://github.com/hanami/router/blob/unstable/spec/integration/hanami/router/mount_spec.rb)
+
+## Mounting an application
+
+In Hanami, you can mount any rack application in desired route. If we want to do it, you should use the `mount` keyword.
```
cllns, line 11: This is in the rack configuration, right? `config.ru` rather than a route files?

swilgosz: oh, is it? I will check, thanks! I know it was in the config in 1.3, but I think I saw it changed in new tests.

Will get back to it! cc: @jodosha

content/v2.0/router/overview.md
swilgosz: Yeah, agree:)

content/v2.0/router/advanced-usage.md
jodosha: Nope, it's the router.

@swilgosz I strongly advice to wrap those lines into the following block:

```ruby
# config/routes.rb
module MyApp
  class ApplicationRoutes < Hanami::Application::Routes
    mount # ...
  end
end
```

Please be aware that this syntax may change, but at least it's crystal clear of what we're talking about: application routes, and not `config.ru`.

---

I would not use "keyword" but "method" instead. Usually keywords are langs constructs.

swilgosz: @jodosha ATM Hanami2APplication template does not inherit from Hanami::Application::Routes, but rather passes a block to the main class

https://github.com/hanami/hanami-2-application-template/blob/main/config/routes.rb

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+
+All the HTTP requests starting with `/sinatra` will be routed to `SinatraApp`.
+
+#### Mounting On A Subdomain
+
+```ruby
+# /config/routes.rb
+
+Hanami.application.routes do
+    mount Blog.new, host: "blog"
```
cllns, line 42:
```suggestion
  mount Blog.new, host: "blog"
```

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+
+Hanami.applicaton.routes do
+  resources :books
+end
+
+```
+
+It generates:
+
+<table class="table table-bordered table-striped">
```
cllns, line 81: Can we change this to (GitHub-flavored) Markdown tables? https://github.github.com/gfm/#tables-extension-

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+    <td>/books/:id</td>
+    <td>Books::Destroy</td>
+    <td>:destroy</td>
+    <td>:book</td>
+  </tr>
+</table>
+
+### Remove Routes
+
+In case we don"t need all the default routes we can use `:only` and pass one or more action names.
```
cllns, line 142:
```suggestion
In case we don't need all the default routes we can use `:only` and pass one or more action names.
```

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+We can also black list routes with `:except`.
```
cllns, line 143:
```suggestion
We can also exclude routes with `:except`.
```

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+Hanami.applicaton.routes do
+  resources :books, controller: "manuscripts"
+
+  # GET /books/1 will route to Manuscripts::Show, etc.
+end
+```
+
+## RESTful Resource
+
+You can also use a `resource` helper, to generate routes for singular resource.
```
cllns, line 196:
```suggestion
You can also use the `resource` helper, to generate routes for a singular resource.
```

content/v2.0/router/advanced-usage.md
```diff
@@ -0,0 +1,624 @@
...
+  <tr>
+    <td>DELETE</td>
+    <td>/account</td>
+    <td>Account::Destroy</td>
+    <td>:destroy</td>
+    <td>:account</td>
+  </tr>
+</table>
+
+### Remove Routes
```
cllns, line 260: There's a lot of duplicated content in this file, from this point onward. Think a couple extra pastes might have gotten saved by accident :)

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,267 @@
+---
+title: "Router: Basic Usage"
+order: 20
+---
+
+So far we've only created a very basic endpoint, but usually, our applications will need much more logic to render the page's content.
+
+## Rack
+
+Hanami is compatible with [Rack SPEC](http://www.rubydoc.info/github/rack/rack/master/file/SPEC), and so the endpoints that we use MUST be compliant as well.
```
cllns, line 10:
```suggestion
Hanami::Router is compatible with [Rack specification](http://www.rubydoc.info/github/rack/rack/master/file/SPEC.rdoc), and so the endpoints that we use MUST be compliant as well.
```

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,267 @@
...
+  module Actions
+    module Home
+      class Index < Hanami::Action
+      end
+    end
+  end
+end
+```
+
+`Home::Index` action of the `Main` slice, will call a corresponding view object with prepared parameters. The view then will render the prepared `slim` template.
```
cllns, line 72: I can understand including _some_ info about Actions in the Router documentation, since they integrate with each other and it might not be obvious how that happens. I don't think reaching to the next level and talking about views in the router documentation is a good idea though.

content/v2.0/router/basic-usage.md
cllns: I'm not Tim but I think this is a good approach. There's a `slice` method in the router, so that's what's relevant, but any more info (about how slices themselves work) should go in the slices documentation :)

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,267 @@
...
+# /config/routes.rb
+
+Hanami.application.routes do
+  get "/books/:book_id/reviews/:id", to: "reviews.show"
+end
+```
+
+### Variables Constraints
+
+It"s possible to specify constraints for each variable.
```
cllns, line 173:
```suggestion
It's possible to specify constraints for each variable.
```

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,267 @@
...
+The rule MUST be expressed as a regular expression.
+If a request can satisfy all of them, we"re good, otherwise, a `404` is returned.
```
cllns, line 175:
```suggestion
If a request can satisfy all of them, we're good, otherwise, a `404` is returned.
```

cllns: Ah, I guess you replaced `'` with `"` throughout, I see now :)

content/v2.0/router/basic-usage.md
```diff
@@ -0,0 +1,267 @@
...
+  end
+end
+
+# This will generate:
+#
+#   /docs/installation
+#   /docs/usage
+```
+
+That's it for basic usage. If you're keen to expertise this part. Check out the next chapter!
```
cllns, line 267:
```suggestion
That's it for basic usage. If you're keen to learn more, check out the advanced usage documentation.
```

content/v2.0/router/overview.md
```diff
@@ -0,0 +1,75 @@
...
+
+```ruby
+# slices/main/actions/home/index.rb
+
+module Main
+  module Actions
+    module Home
+      class Index < Main::Action
+        def handle(req, res)
+          res.body = "<h1>Hello, Hanami!</h1>"
```
cllns, line 67:
```suggestion
          res.body = "Hello, Hanami!"
```
This could be HTML but it could also be plaintext (or JSON, or XML, or anything). I don't see writing HTML here as being all that helpful, and I see it potentially being confusing (since the action doesn't care what it its body is, to some extent).

content/v2.0/router/advanced-usage.md
swilgosz: Sure. I copied it from existing docs, but if other syntax works better, I don't mind.

content/v2.0/router/basic-usage.md
swilgosz: Oh, yeah I've got used to single quotes in the code, but hanami developers prefer double quotes so I wanted to unify to keep things tied together :). Too quick. Will look for the rest of overlooks, thanks for the review!

content/v2.0/architecture/overview.md
```diff
@@ -0,0 +1,206 @@
...
+> `Hanami::Model` is the default choice for persisting our Ruby objects.
+This is a _soft-dependency_, it can be removed from our `Gemfile` and replaced with something else.
+
+Let's have a look at how the `lib/` directory appears for a new generated project called `bookshelf` that uses `Hanami::Model`.
+
+```shell
+lib
+├── bookshelf
+│   ├── action.rb
+│   ├── entities
```
svoop, line 36: @swilgosz No biggie, but maybe a tailing slash for directories would improve the readability, e.g.

```
├── entities/
├── entities.rb
├── foodir/
    └── barfile
```

swilgosz: good idea! BTW: not sure, if you are aware, but you can suggest changes inside reviews, that I can accept directly as your commit

svoop: @swilgosz IC, will do next time.

swilgosz:
Closed in favour of #97  #98 #99  and other per-chapter prs

PR #91, opened 2021-06-09
Add note about Hanami v1.3 requiring Ruby 2

cllns:
We should tell users going through the Getting Started guide that Hanami v1.3 doesn't support Ruby 3.

See: 
- https://github.com/hanami/hanami/issues/1108
- https://github.com/hanami/hanami/issues/1101

I also created a PR to require Ruby 2 for Hanami v1.3: https://github.com/hanami/hanami/pull/1109


adam12 (commented):
Can you rebase against `master`? This file doesn't exist in that branch I don't believe, as everything was moved under the `v1.3` namespace.

cllns:
Oops, right! Rebased off `master` and force-pushed the commit, so it should be good now.

adam12 (approved):

PR #93, opened 2021-06-26
Update getting started guide

cllns:
The main change I wanted to make here was that I realized our spec [where we introduce dynamic data was a false positive](https://github.com/hanami/guides/blob/main/content/v1.3/introduction/getting-started.md#displaying-dynamic-data). That's a bad practice for Behavior Driven Development, so I changed the spec to be a single book, and check for the actual content, rather than a CSS selector and count. This is more robust.

I also changed the link from "Clean architecture" to "Hexagonal architecture". The latter has a wikipedia page and it gets the underlying idea across well enough. (Plus we want to distance ourselves from Robert Martin, who drafted the post we are currently linking too.)

In the process, I went through the whole guide and cleaned up many pretty small things.

### DIFF
--- a/content/v1.3/introduction/getting-started.md
+++ b/content/v1.3/introduction/getting-started.md
@@ -923,7 +923,7 @@ To make our tests pass, we need to implement validations.
 Although you can add validation rules to the entity, Hanami also allows you to define validation rules as close to the source of the input as possible, i.e., the action.
 Hanami controller actions can use the `params` class method to define acceptable incoming parameters.
 
-This approach both whitelists what `params` are used (others are discarded to prevent mass-assignment vulnerabilities from untrusted user input) _and_ adds rules to define what values are acceptable — in this case, we've specified that the nested attributes for a book's title and author should be present.
+This approach both explicitly declares which `params` are allowed (others are discarded to prevent mass-assignment vulnerabilities from untrusted user input) _and_ it adds rules to define which values are acceptable — in this case, we've specified that the nested attributes for a book's title and author should be non-empty strings.
 
 With our validations in place, we can limit our entity creation and redirection to cases where the incoming `params` are valid:
 
--- a/content/v1.3/introduction/getting-started.md
+++ b/content/v1.3/introduction/getting-started.md
@@ -296,12 +296,12 @@ This test means that when we go to [/books](http://localhost:2300/books),
 we'll see two HTML elements that have class `book`,
 and both will be inside of an HTML element that has an id of `books`.
 
-Our test suite is `Unable to find visible css "#books"`.
+Our test suite shows 1 failure: `Unable to find visible css "#books"`.
 
 Not only are we missing that element,
 we don't even have a page to put that element on!
 
-Let's create a new action to fix that.
+Let's create a new action and a new route to fix that.
 
 ### Hanami Generators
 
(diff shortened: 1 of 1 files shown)


content/v1.3/introduction/getting-started.md
```diff
@@ -1133,9 +1137,22 @@ We can use the `routes` helper method that is available in our views and actions
...
 
+and in `apps/web/templates/books/index.html.erb`:
+
+```erb
+...
+<a href="<%= routes.new_book_path %>">New book</a>
+...
+```
+
+(You could also extend this to the specs if you'd like,
```
adam12, line 1153: Do you think we need to parenthesis this statement?

content/v1.3/introduction/getting-started.md
```diff
@@ -491,12 +493,15 @@ We can verify it all works as expected with a unit test:
...
   it 'can be initialized with attributes' do
-    book = Book.new(title: 'Refactoring')
+    book = Book.new(title: 'Refactoring', author: 'Martin Fowler')
     expect(book.title).to eq('Refactoring')
+    expect(book.author).to eq('Martin Fowler')
   end
 end
 ```
 
+(Generally we recommend against 'testing the framework' like this, but it's useful here to demonstrate how Entities work in Hanami.)
```
adam12, line 503: I'm not sure I feel this needs to be parenthesized.

content/v1.3/introduction/getting-started.md
```diff
@@ -549,8 +554,10 @@ RSpec.describe 'List books' do
 end
 ```
 
-We create the required records in our test and then assert the correct number of book classes on the page.
-When we run this test, it should pass. If it does not pass, a likely reason is that the test database was not migrated.
+We create a single Book record in our test and then assert that the title and author are displayed on the page.
+When we run this test, it will fail.
+This is because we haven't updated our template yet,
+it still has the hard-coded books from earlier.
```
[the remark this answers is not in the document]
cllns: Updated this to "We create a single Book record in our test and then expect that the title and author are displayed on the page. This test fails since we haven't updated our template yet, and it still includes the hard-coded books from earlier."

What do you think of that?

content/v1.3/introduction/getting-started.md
cllns: Removed parenthesis, and added _why_ that might be prefererd.

"You could also extend this to the specs if you'd like, but changing the path can affect users so you may want to repeat yourself there (to make a potentially breaking change for users harder to implement)."

I don't _love_ that this is the note we end on before the "Wrapping Up" section, since it's not directly related to Hanami.

content/v1.3/introduction/getting-started.md
cllns: I changed this to a "convention" block, which seems appropriate. It's a (Green) aside now, which feels right. Users can test the framework if they want! I just wouldn't :)

cllns:
Thanks for the careful and thoughful review @adam12 :)

I think I addressed the concerns. Let me know about the items I left "open" (instead of resolved). Open to more tweaks too

cllns:
@adam12 Can you check this again when you get a chance? I added a couple other small commits

content/v1.3/introduction/getting-started.md
adam12: > It's a (Green) aside now, which feels right.

I'm not seeing this in the preview. Was this something you saw locally?

adam12 (approved):

content/v1.3/introduction/getting-started.md
cllns: Whoops, I had a typo :) https://github.com/hanami/guides/pull/93/commits/05d5f5c413ef228802eda843d18c8fb2fa61d7cd

cllns:
Found and [fixed](https://github.com/hanami/guides/commit/7fadff3788d10be060564bdeeb846e5cc6d4cc3b) an error post-merge 🙃

PR #102, opened 2022-02-09
Add `BRANCH` env variable to allowed getenv params

adam12:
One of the newer Hugo releases has increased security around the `getenv` function, and it prevents the callout to the `BRANCH` env variable on newer releases of Hugo, preventing builds.


cllns (approved):
I ran into this too, so thanks!

PR #106, opened 2022-03-11
Main style to SCSS, additional style adjustments

baweaver:
This change introduces SCSS into the pipeline to simplify and
consolidate styles into a more hierarchical structure. It also
eliminates redundant or unused styles.

Additional changes have been made to normalize colors, add hover-effect
differences, change font sizes, change inline code styles, and other
minor changes.

After and Before image:

image

### DIFF
--- a/themes/hanami/assets/scss/main.scss
+++ b/themes/hanami/assets/scss/main.scss
@@ -1,0 +1,317 @@
+// Variables
+
+$color-cyan: #35DDFF;
+$color-purple: #685D9F;
+$color-orange: #fc7c5f;
+$color-red: #FF6886;
+$color-mid-red: #ae2323;
+$color-green: #46E895;
+$color-yellow: #FFD842;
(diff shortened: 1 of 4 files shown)


jodosha (changes requested):
@baweaver Thanks for taking care of this, but there are two main problems with this PR.

## Deployment is broken

See the PR [preview](https://deploy-preview-106--hanami-guides.netlify.app/) and the attached screenshot.

Steps to reproduce:

1. From the root of the project: run `hugo`
2. `cd public`
3. `ruby -run -e httpd . -p 5000`
4. Visit `http://localhost:5000`
homepage-deploy

## Colors are opaque

Steps to reproduce:

1. From the root of the project: run `bin/server`
3. Visit `http://127.0.0.1:1313`

Here's the before (left) and after (right) of the home page and getting started article
homepage

gettingstarted

PR #108, opened 2022-04-30
Document compatible hugo version

### LINKED ISSUE #103: Force Hugo version for Netlify
Hugo seems to move fairly quickly and occasionally breaks backwards compatibility.

The version of Hugo we're using to build the site on Netlify and locally is ambiguous. If you're using Homebrew locally it's likely the most recent release of Hugo, but Netlify uses the [first version the site was ever built with](https://docs.netlify.com/configure-builds/common-configurations/hugo/#hugo-version).

We should define the version in the README and in a Netlify configuration file to be consistent.

tak1n:
Closes #103

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -34,4 +34,4 @@ pygmentsstyle = "monokailight"
   getenv = ['^HUGO_', 'BRANCH']
 
 [build.environment]
-  HUGO_VERSION = "v0.92.3"
+  HUGO_VERSION = "v0.92.2"
--- a/config.toml
+++ b/config.toml
@@ -32,6 +32,3 @@ pygmentsstyle = "monokailight"
 
 [security.funcs]
   getenv = ['^HUGO_', 'BRANCH']
-
-[build.environment]
-  HUGO_VERSION = "v0.92.2"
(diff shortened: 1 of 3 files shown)


config.toml
```diff
@@ -32,3 +32,6 @@ pygmentsstyle = "monokailight"
 
 [security.funcs]
   getenv = ['^HUGO_', 'BRANCH']
+
+[build.environment]
+  HUGO_VERSION = "v0.92.2"
```
adam12, line 37: I'm not sure this is the correct file to place this. Do you remember where you sourced it from? I presumed it might be [`netlify.toml`](https://docs.netlify.com/integrations/frameworks/hugo/#hugo-version) or something.

adam12 (commented):
Thanks for tackling this. Only had a question about the Netlify config.

config.toml
tak1n: TBH, I'm not sure either. Should I move it to `netlify.toml`? Also not sure how we can double check which hugo version is really used on netlify then.

adam12:
Thanks for the PR!

I'm a little suspect of the `v` in the version string, but Netlify didn't complain so perhaps it strips it for us.

PR #109, opened 2022-05-20
Remove text-align: justify; from main p

emp823:
There are some text spacing issues in the guides that I found distracting. `text-align: justify` seems to be the culprit. I'm not sure what the original intent for the style was, so this change may have consequences elsewhere that I'm not aware of.

**GIF**
screencast 2022-05-20 15-43-13


cllns:
Thank you! I have this change (and others) included in #82, which has a (non-trivial) merge conflict due to other changes, unfortunately. But you're right, this definitely should be fixed!

PR #115, opened 2022-11-09
Ensure version switcher to land to Getting Started page

jodosha:
Please manual test with Netifly preview.

https://trello.com/c/93hWs3Rm


andrewcroome:
Looking at the HTML, it seems a double slash has crept in @jodosha. 

Screen Shot 2022-11-10 at 11 53 06 am

Drenmi:
@jodosha I was thinking if we should alias `/1.3` and `/2.0` in the introduction pages' front matter. That would serve to make e.g. `guides.hanamirb.org/v2.0` a valid entry point as well.

It could be done in addition to the change in this PR, or instead of it. WDYT?

jodosha:
@andrewcroome Good catch. Fixed in https://github.com/hanami/guides/pull/115/commits/f57e333bb42dbe005fad4bd9b405282da337bf1b

Screenshot 2022-11-10 at 14 32 46

jodosha:
@Drenmi Not in this PR.

Reasons:

* Never mix fixes with enhancements
* I don't have time to create main pages for minor versions

🙂

Drenmi:
> I don't have time to create main pages for minor versions

This is a good point I did not think of, but it does explain why it felt a bit weird that this would go in the front-matter of specific pages. 🙂  If I figure out some way to do this as a general rule in Hugo I'll make a PR.

PR #116, opened 2022-11-13
Add 2.0 getting started walk through.

andrewcroome:
This adds a getting started walk though for 2.0, similar in approach to the 1.3 guide.

This is ready for review and feedback.


content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+Blocks are convenient, but let's adjust our root route to invoke an action instead:
+
+```ruby
+# config/routes.rb
+
+# frozen_string_literal: true
+
+module Bookshelf
+  class Routes < Hanami::Routes
+    root to: "home.index"
```
timriley, line 288: Should this actually be `"home.show"`? The home page is a singleton resource and we're showing it, rather than providing a list of home pages.

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+  start do
+    @config.auto_registration(
+      target.root.join("lib/bookshelf/persistence"),
+      namespace: "Bookshelf::Persistence"
+    )
+
+    register "rom", ROM.container(@config)
+  end
+end
+```
```
timriley, line 565: Feels like we could _probably_ get away without:

`require "rom-changeset"`

`config.plugin(:sql, relations: :auto_restrictions)`

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+```ruby
+# frozen_string_literal: true
+
+Hanami.app.register_provider :persistence, namespace: true do
+  prepare do
+    require "rom-changeset"
+    require "rom/core"
+    require "rom/sql"
+
+    @config = ROM::Configuration.new(:sql, target["settings"].database_url)
```
timriley, line 548: We're setting `@config` here but using `config` in the lines immediately below. This should be `@config` in those places, right? I realised this is copied from decafsucks, so I think this is an issue to be fixed there too.

I just double checked and the only reason this actually works is because providers include `Dry::Configurable`, which has its _own_ `config` method that reads from a memoized `@config` ivar 😬

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+end
+```
+
+And then append the following line to `spec/spec_helper.rb`:
+
+```ruby
+require_relative "support/database_cleaner"
+```
+
+Finally, enable rom-rb's rake tasks for database migrations by appending the following to the `Rakefile`:
```
timriley, line 683: It might be good to add another callout box at the end of this section just reminding the user that all of these manual steps will not be required as of Hanami 2.1, which will be released in early 2023.

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+    ])
+  end
+end
+```
+
+To get this spec to pass, we'll need to update our books index action to return books from the books relation.
+
+To access the books relation within the action, we can use Hanami's "Deps mixin". Covered in detail in the [containers and components](/v2.0/architecture/containers/) section of the [Architecture guide](/v2.0/architecture/containers/), the Deps mixin gives each of your application's components easy access to the other components it depends on to achieve its work. We'll see this in more detail in a moment.
+
+For now however, it's enough to know that we can use `include Deps["persistence.rom"]` to make rom-rb available via a `rom` method within our action. The books relation is then available via `rom.relations[:books]`.
```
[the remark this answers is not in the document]
timriley: Ah, I see you have a note a bit further down. Perhaps you can mention there that Hanami 2.1 will have repositories out of the box.

timriley (approved):
It's so great, @andrewcroome!!

Having the rom-rb setup does add a bit of extra mental burden, it's true, but I think we get some of that back by not having to contrive a totally artificial example for the walkthrough, and instead cover something that should feel utterly familiar to most of our readers. I think this is a worthwhile tradeoff.

Left a couple of minor notes in this and the previous review (which I submitted early by accident), but on the whole this is great and I'd be very happy to see it merged.

(The fewer open PRs we have on these guides the better IMO — so long as they're given at least one pass by a core team member, I think we should get them in, and we can always continue to review and tweak post-merge).

content/v2.0/introduction/getting-started.md
```diff
@@ -58,25 +57,27 @@ To create a Hanami application, you will need Ruby 3.0 or greater. Check your ru
...
 
 In order to create a Hanami application, first install the hanami gem:
 
 ```shell
-gem install hanami --pre
+gem install hanami
 ```
 
-### Generating your first application
+### Using the hanami new command
```
katafrakt, line 70: This reads a bit weird IMO. Can "hanami new" be in backticks, so it's clear that it is the complete command?

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+### Adding persistence using rom-rb
+
+Let's add just enough rom-rb to get persistence working using Postgres.
+
+First, add these gems to the Gemfile and run `bundle install`:
+
+```ruby
+# Gemfile
+gem "rom"
+gem "rom-sql"
```
katafrakt, line 516: The code in guide will work only with `3.6.0` (as of now) - does not work with 3.5.0 IIRC (or rather, it won't resolve to 3.5.0) and definitely won't work with 4.0. Perhaps it's worth to pin the version?

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+
+```ruby
+# Rakefile
+
+require "rom/sql/rake_task"
+
+namespace :db do
+  task setup: :environment do
+    Hanami.app.prepare(:persistence)
+    ROM::SQL::RakeSupport.env = ROM.container(Hanami.app["persistence.config"])
```
katafrakt, line 693: Personally I use

```ruby
    require "rom/core"
    rom_config = ROM::Configuration.new(:sql, Hanami.app["settings"].database_url)
    ROM::SQL::RakeSupport.env = rom_config
```

I don't want to load whole persistence in rake task, because in case the code is misaligned with the database, I cannot ruby `rake db:reset`. But perhaps someone wiser could confirm if this does not cause some issues.

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+      { "title" => "Refactoring", "author" => "Martin Fowler" },
+      { "title" => "Test Driven Development", "author" => "Kent Beck" }
+    ])
+  end
+end
+```
+
+To get this spec to pass, we'll need to update our books index action to return books from the books relation.
+
+To access the books relation within the action, we can use Hanami's "Deps mixin". Covered in detail in the [containers and components](/v2.0/architecture/containers/) section of the [Architecture guide](/v2.0/architecture/containers/), the Deps mixin gives each of your application's components easy access to the other components it depends on to achieve its work. We'll see this in more detail in a moment.
```
katafrakt, line 786: There are two links pointing to the same location. Maybe the first one should point to the exact section and the latter is not necessary?

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+
+For now however, it's enough to know that we can use `include Deps["persistence.rom"]` to make rom-rb available via a `rom` method within our action. The books relation is then available via `rom.relations[:books]`.
+
+To satisy our spec, we need to meet a few requirements. Firstly, we want to render each book's _title_ and _author_, but not its _id_. Secondly we want to return books alphabetically by title. We can achieve these requirements using the `select` and `order` methods offered by the books relation:
```
katafrakt, line 790:
```suggestion
To satisfy our spec, we need to meet a few requirements. Firstly, we want to render each book's _title_ and _author_, but not its _id_. Secondly we want to return books alphabetically by title. We can achieve these requirements using the `select` and `order` methods offered by the books relation:
```

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1261 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+GET /books
+  returns a list of books
+
+Finished in 0.05765 seconds (files took 1.36 seconds to load)
+1 example, 0 failures
+```
+
+## Parameter validation
+
+Of course, returning _every_ book in the database when a visitor makes a request to `/books` is not going to be a good strategy for very long. Luckily rom-rb relations offer pagination support. Let's add pagination with a default page length of 5:
```
katafrakt, line 834: Is it "page length"? I think I usually see "page size".

katafrakt (commented):
some comments from a bystander

content/v2.0/introduction/getting-started.md
andrewcroome: Oh yep, adjusted to "home.show", thanks.

content/v2.0/introduction/getting-started.md
andrewcroome: Seems we certainly can!

content/v2.0/introduction/getting-started.md
andrewcroome: Computers are amazing. Fixed - thank you 😆

content/v2.0/introduction/getting-started.md
andrewcroome: Done.

content/v2.0/introduction/getting-started.md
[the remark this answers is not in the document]
andrewcroome: Thanks - yes, I've added.

content/v2.0/introduction/getting-started.md
andrewcroome: Thanks @katafrakt - I've changed it to "Using the application generator". I did try the backticks but the formatting doesn't look great in a heading.

content/v2.0/introduction/getting-started.md
andrewcroome: Ah, right - thank you. I've gone with:

```
gem "rom", "~> 5.3"
gem "rom-sql", "~> 3.6"
```

content/v2.0/introduction/getting-started.md
andrewcroome: Oh yep - thanks @katafrakt. I've gone with your version.

content/v2.0/introduction/getting-started.md
andrewcroome: Yeah, that makes more sense - adjusted to page size.

content/v2.0/introduction/getting-started.md
katafrakt: nice, that's probably even better heading

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1243 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+
+Actions define a `#handle` method which accepts a `request` object, representing the incoming request, and a `response` object, representing the outgoing response.
+
+```ruby
+def handle(request, response)
+  # ...
+end
+```
+
+In the generated action above, `*` is used for the `request` argument because the action does not currently use the request.
```
solnic, line 337: The example above doesn't have `*` in the signature

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1243 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+
+```ruby
+# frozen_string_literal: true
+
+RSpec.describe "GET /books", type: [:request, :database] do
+  let(:books) { app["persistence.rom"].relations[:books] }
+
+  before do
+    books.insert(title: "Refactoring", author: "Martin Fowler")
+    books.insert(title: "Test Driven Development", author: "Kent Beck")
```
solnic, line 750: It'd be really nice to have a non-male author as an example too. Maybe "POODR" by Sandi Metz?

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1243 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+# frozen_string_literal: true
+
+module Bookshelf
+  module Actions
+    module Books
+      class Index < Bookshelf::Action
+        include Deps["persistence.rom"]
+
+        params do
+          optional(:page).filled(:integer, gt?: 0)
```
solnic, line 918:
```suggestion
          optional(:page).value(:integer, gt?: 0)
```

content/v2.0/introduction/getting-started.md
```diff
@@ -138,29 +162,1243 @@ Visit your application in the browser at [http://localhost:2300](http://localhos
...
+          optional(:per_page).filled(:integer, gt?: 0, lteq?: 100)
```
solnic, line 919:
```suggestion
          optional(:per_page).value(:integer, gt?: 0, lteq?: 100)
```

content/v2.0/introduction/getting-started.md
andrewcroome: Done, thanks @solnic.

content/v2.0/introduction/getting-started.md
andrewcroome: Ta. I've made this more clear hopefully.

content/v2.0/introduction/getting-started.md
andrewcroome: Updated, thank you.

PR #101, opened 2022-01-17
Guides 2.0 - Actions

swilgosz:
Overview
Actions explanation for guides 2.0

- [x] Overview
- [x] Basic Usage
- [x] Parameters
- [x] [Request & Response](https://github.com/hanami/guides/pull/117/files)
- [ ] Exposures - 2.1
- [x] [Rack Integration](https://github.com/hanami/guides/pull/120)
- [x] MIME Types
- [x] Cookies
- [x] Sessions
- [x] Exception Handling
- [x] Control Flow
- [ ] HTTP Caching
- [ ] Share Code
- [ ] Testing

PR #121, opened 2022-11-16
Guides 2.0 Actions # parameters

swilgosz:
Resolves: https://trello.com/c/Cxry5P0K/302-actions-parameters

Extracted from: https://github.com/hanami/guides/pull/101

### DIFF
--- a/content/v2.0/actions/overview.md
+++ b/content/v2.0/actions/overview.md
@@ -62,8 +62,6 @@ Topics to cover:
 - request and response
 - mime types and format
 - control flow
-- params (including req.ar.to_h) and query string parameters.
-- param validation
 - request headers
 - response headers
 - redirects
--- a/content/v2.0/actions/overview.md
+++ b/content/v2.0/actions/overview.md
@@ -14,18 +14,18 @@ Every action in your Hanami application is an individual class. Actions define a
 Modifying the response object allows you to control how your application responds to a request.
 
 ```ruby
-# app/actions/books/show.rb
+# app/actions/home/show.rb
 
 # frozen_string_literal: true
 
 module Bookshelf
   module Actions
-    module Books
+    module Home
       class Show < Bookshelf::Action
         def handle(request, response)
-          id = request.params[:id]
+          name = request.params[:name]
 
-          response.body = "You requested book #{id}"
+          response.body = "Welcome to Bookshelf #{name}!"
         end
       end
     end
@@ -33,7 +33,10 @@ module Bookshelf
 end
 ```
 
-Let's take a deeper look.
+As the code above suggests, the `request` object provides access to the parameters associated with the incoming request through a `#params` method.
+
+Let's start by taking a look at action [parameters](/v2.0/actions/parameters/).
+
 
 ## Parameters
 
--- a/content/v2.0/actions/parameters.md
+++ b/content/v2.0/actions/parameters.md
@@ -1,11 +1,10 @@
 ---
 title: Parameters
 order: 30
-aliases:
-  - "/actions/parameters"
 ---
 
-Parameters are taken from the Rack env and passed as an argument to `#call`.
+Parameters are taken from the Rack env, wrapped into the `Params` object, and passed as an argument to `#handle` method.
+
 They are similar to a Ruby `Hash`, but they offer an expanded set of features.
 
 ## Sources
@@ -23,11 +22,11 @@ To access the value of a param, we can use the _subscriber operator_ `#[]`.
 ```ruby
 # app/controllers/dashboard/index.rb
 
-module Sandbox
-  module Controllers
-    module Dashboard
+module Bookshelf
+  module Actions
+    module Books
       class Index < Action
-        def def handle(req, res)
+        def handle(req, res)
           req.body = "Query string: #{ req.params[:q] }"
         end
       end
@@ -76,7 +75,8 @@ We use `.params` to map the structure of the (nested) parameters.
 
 ```ruby
 # app/actions/signup/create.rb
-module Sandbox
+
+module Bookshelf
   module Actions
     module Signup
       class Create < Action
@@ -158,7 +158,7 @@ We can coerce the Ruby type, validate if a param is required, determine if it is
 
 ```ruby
 # app/actions/signup/create.rb
-module Sandbox
+module Bookshelf
   module Actions
     module Signup
       class Create < Action
@@ -197,10 +197,10 @@ An alternative is to extract a class and pass it as an argument to `.params`.
 ```ruby
 # app/actions/signup/my_params.rb
 
-module Web
-  module Controllers
+module Bookshelf
+  module Actions
     module Signup
-      class MyParams < Web::Action::Params
+      class MyParams < Hanami::Action::Params
         MEGABYTE = 1024 ** 2
 
         params do
@@ -218,8 +218,7 @@ end
 ```
 
 ```ruby
-# apps/web/controllers/signup/create.rb
-require_relative './my_params'
+# apps/actions/signup/create.rb
 
 module Sandbox
   module Actions
@@ -239,44 +238,6 @@ module Sandbox
   end
 end
 ```
-
-### Inline predicates
-
-In case there is a predicate that is needed only for the current params, you can define inline predicates:
-
-```ruby
-# app/actions/books/create.rb
-
-module Sandbox
-  module Actions
-    module Books
-      class Create < Action
-        params Class.new(Hanami::Action::Params) {
-          predicate(:cool?, message: "is not cool") do |current|
-            current.match(/cool/)
-          end
-
-          validations do
-            required(:book).schema do
-              required(:title) { filled? & str? & cool? }
-            end
-          end
-        }
-
-        def handle(req, res)
-          if req.params.valid?
-            res.body = 'OK'
-          else
-            res.body = params.error_messages.join("\n")
-          end
-        end
-      end
-    end
-  end
-end
-```
-
-
 
 ## Body Parsers
 
@@ -312,9 +273,11 @@ $ curl http://localhost:2300/books      \
 To properly coerce `json` params, you should enable this feature.
 
 ```ruby
-#config.ru
-require "hanami/middleware/body_parser"
-use Hanami::Middleware::BodyParser, :json
+# config/app.rb
+
+class App < Hanami::App
+  config.middleware.use :body_parser, :json
+end
 ```
 
 Now `params.get(:book, :title)` returns `"Hanami"`.
@@ -339,6 +302,7 @@ and subsequently register it:
 ```ruby
 # /config.ru
 
-require "hanami/middleware/body_parser"
-use Hanami::Middleware::BodyParser, FooParser.new
+class App < Hanami::App
+  config.middleware.use :body_parser, FooParser.new
+end
 ```
--- a/content/v2.0/actions/parameters.md
+++ b/content/v2.0/actions/parameters.md
@@ -81,11 +81,11 @@ module Bookshelf
     module Signup
       class Create < Action
         params do
-          required(:email).filled(:str?)
-          required(:password).filled(:str?)
+          required(:email).filled(:string)
+          required(:password).filled(:string)
     
           required(:address).schema do
-            required(:country).filled(:str?)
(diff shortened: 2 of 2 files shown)


content/v2.0/actions/overview.md
```diff
@@ -62,8 +62,6 @@ Topics to cover:
 - request and response
 - mime types and format
 - control flow
-- params (including req.ar.to_h) and query string parameters.
```
swilgosz, line 65: What is `req.ar` ? This method doesn't seem to be defined.

swilgosz:
[Actions: parameters](https://trello.com/c/Cxry5P0K/302-actions-parameters)

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+To access the value of a param, we can use the _subscriber operator_ `#[]`.
+
+```ruby
+# app/controllers/dashboard/index.rb
+
+module Bookshelf
+  module Actions
+    module Books
+      class Index < Action
+        def handle(req, res)
```
solnic, line 29: We're using `request`, and `response` after all so it should be used here as well.

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+Because `params[:book]` is `nil`, we can't access `:title`.
+In this case Ruby will raise a `NoMethodError`.
+
+We have a safe solution for our problem: `#get`.
+It accepts a list of symbols, where each symbol represents a level in our nested structure.
+
+```ruby
+params.get(:book, :title)             # => "Hanami"
+params.get(:unknown, :nested, :param) # => nil instead of NoMethodError
+```
```
solnic, line 59: Oh noez, I had no idea this exists. Too late now but...maybe not? If we also provide `dig` then please do change it to `dig` here, if not...then oh well, something we can address in 3.0.

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+  module Actions
+    module Signup
+      class Create < Action
+        params do
+          required(:email).filled(:str?)
+          required(:password).filled(:str?)
+    
+          required(:address).schema do
+            required(:country).filled(:str?)
+          end
```
solnic, line 89: Please don't use type predicates and switch to type specs. So `:str?` becomes `:string` etc. This is how it's going to work in dry-schema 2.0 by default and it will be a mandatory argument.

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+
+Imagine we introduce a second feature: _"Invitations"_.
+An existing user can ask someone to join.
+Because the invitee will decide a password later on, we want to persist that `User` record without that value.
+
+If we put `password` validation in `User`, we need to handle these two use cases with a conditional.
+But in the long term this approach is painful from a maintenance perspective.
+
+```ruby
+# Example of poor style for validations
```
solnic, line 121: I would drop this comment. In general we should avoid presenting other solutions in a negative way. I used to do this a lot thinking that I'm just explaining why I do X when something else does Y but it actually makes things worse as people may get upset when what they enjoy is being presented as "worse" or "poor style" etc.

It should be sufficient to explain why certain things work the way they work and the reasoning behind it without contrasting it with other solutions so often.

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+  private
+  
+  def password_required?
+     !invited_user? && !admin_password_reset?
+  end
+end
+```
+
+We can see validations as the set of rules for data correctness that we want for **a specific use case**.
+For us, a `User` can be persisted with or without a password, **depending on the workflow** and the route through
```
solnic, line 134:
```suggestion
For us, a `User` can be persisted with or without a password, **depending on the use case** and the route through
```

I think it's fine to repeat the same word here as `workflow` has specific meanings in various contexts (ie background job workflow etc.).

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+which the `User` is persisted.
+
+### Boundaries
+
+The second important aspect is that we use validations to prevent invalid inputs to propagate in our system.
+In an MVC architecture, the model layer is the **farthest** from the input.
+It's expensive to check the data right before we create a record in the database.
+
+If we **consider correct data as a precondition** before starting our workflow, we should stop unacceptable inputs as soon as possible.
```
solnic, line 143:
```suggestion
If we **consider correct data as a precondition** before starting our workflow, we should stop invalid inputs as soon as possible.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+        MEGABYTE = 1024 ** 2
+
+        params do
+          required(:name).filled(:str?)
+          required(:email).filled(:str?, format?: /@/)
+          required(:password).filled(:str?)
+          required(:terms_of_service).filled(:bool?)
+          required(:age).filled(:int?, included_in?: 18..99)
+          optional(:avatar).filled(size?: 1..(MEGABYTE * 3))
+        end
```
solnic, line 174: All of this needs to be updated to use type specs, `:str?` -> `:string` `:bool?` -> `:bool` `:int?` -> `:integer` and avatar should be `:string` before the size? check

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,308 @@
...
+  end
+end
+```
+
+Parameter validations are delegated, under the hood, to [Hanami::Validations](https://github.com/hanami/validations).
+Please check the related documentation for a complete list of options and how to share code between validations.
+
+## Concrete Classes
+
+The params DSL is really quick and intuitive but it has the drawback that it can be visually noisy and makes it hard to unit test.
```
solnic, line 194: This could be explained better. ie Why does it make things harder to unit test? I would also drop "unit test" and just say "test".

solnic (changes requested):
Thanks for working on this 👏🏻 I've left some suggestions / questions.

content/v2.0/actions/parameters.md
swilgosz: Addressed.

content/v2.0/actions/parameters.md
swilgosz: Both work the same - replaced with `dig`. We can remove the `get` if no needed in the upcoming releases.

content/v2.0/actions/parameters.md
swilgosz: done

content/v2.0/actions/parameters.md
swilgosz: Done. Replaced all.

content/v2.0/actions/overview.md
solnic: @swilgosz no such thing, must be some kind of a mistake here

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+- if `per_page` is present, it must be an integer
+
+With this schema in place, a request with a query string of `/books?page=1&per_page=10` will result in:
+
+```ruby
+request.params[:page] # => 1
+request.params[:per_page] # => 10
+```
+
+Note that, due to the `params` block, 1 and 10 have been coerced to integer values (with this they would otherwise be strings).
```
solnic, line 110:
```suggestion
Notice that thanks to the defined params schema with types, `"1"` and `"10"` are coerced to their integer representations `1` and `10`.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+
+Additional rules can be added to apply further contraints. The following params block specifies that, when present, `page` and `per_page` be greater than or equal to 1, and also that `per_page` be less than or equal to 100.
```
solnic, line 112:
```suggestion
Additional rules can be added to apply further constraints. The following params block specifies that, when present, `page` and `per_page` must be greater than or equal to 1, and also that `per_page` must be less than or equal to 100.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+
+```ruby
+params do
+  optional(:page).value(:integer, gte?: 1)
+  optional(:per_page).value(:integer, gte?: 1, lteq?: 100)
+end
+```
+
+Importantly, now that our params block is doing more than just type coercion, we need to explicitly check for and handle our parameters being invalid.
```
solnic, line 121:
```suggestion
Importantly, now that our params schema is doing more than just type coercion, we need to explicitly check for and handle our parameters being invalid.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+
+module Bookshelf
+  module Actions
+    module Users
+      class Create < Bookshelf::Action
+        params do
+          required(:email).filled(:string)
+          required(:password).filled(:string)
+
+          required(:address).schema do
```
solnic, line 167:
```suggestion
          required(:address).hash do
```

plain `schema` doesn't specify the type to be `:hash`, which is what `hash` macro does.

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+    end
+  end
+end
+```
+
+The `params` block in this action specifies that:
+
+- `email`, `password` and `address` parameters are required to be present.
+- `address` has `street` and `country` as nested parameters, which are also required.
+- each of `email`, `password`, `street` and `country` must be filled (non-blank) strings.
```
solnic, line 192:
```suggestion
- `email`, `password`, `street` and `country` must be filled (non-blank) strings.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+module Bookshelf
+  module Actions
+    module Users
+      module Params
+        class Create < Hanami::Action::Params
+          params do
+            required(:email).filled(:string)
+            required(:password).filled(:string)
+
+            required(:address).schema do
```
solnic, line 249:
```suggestion
            required(:address).hash do
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+}
+```
+
+Action validations use the [dry-validation](https://dry-rb.org/gems/dry-validation/) gem, which provides a powerful DSL for defining schemas.
+
+Consult the [dry-validation](https://dry-rb.org/gems/dry-validation/) and [dry-schema](https://dry-rb.org/gems/dry-schema/) gems for further documentation.
+
+## Using concrete classes
+
+In addition to specifying parameter validations "inline" in a `params` block, actions can also hand their validation responsibilities to a separate class.
```
solnic, line 231:
```suggestion
In addition to specifying parameter validations "inline" in a `params` block, actions can also hand over their validation responsibilities to a separate class.
```

content/v2.0/actions/parameters.md
```diff
@@ -0,0 +1,350 @@
...
+end
+```
+
+## Validations at the HTTP layer
+
+Validating parameters in actions is useful for validating parameter structure, performing parameter coercion and type validations.
+
+More complex domain-specific validations, or validations concerned with things such as uniqueness, however, are usually better performed at layers deeper than your HTTP actions.
+
+For example, verifying a that an email address has been provided is something an action parameter validation should reasonably do, but checking that a user with that email address doen't already exist is unlikely to be a good responsibility for a HTTP action to have. That validation might instead be peformed by a create user operation, which can perform a check against a user store.
```
solnic, line 285:
```suggestion
For example, verifying that an email address has been provided is something an action parameter validation should reasonably do, but checking that a user with that email address doesn't already exist is unlikely to be a good responsibility for an HTTP action to have. That validation might instead be performed by a create user operation, which can perform a check against a user store.
```

solnic (changes requested):

content/v2.0/actions/parameters.md
swilgosz: I agree. Official docs should NOT be subjective. I can present things filled in with more personal opinion in un/semi-official channels like Hanami Mastery.

andrewcroome (approved):

PR #122, opened 2022-11-19
Add settings guide

andrewcroome:
This is adds a section on settings within an application guide.

Previewable [here](https://deploy-preview-122--hanami-guides.netlify.app/v2.0/application/settings/)

### DIFF
--- a/content/v2.0/application/settings.md
+++ b/content/v2.0/application/settings.md
@@ -333,3 +333,9 @@ Exactly which `.env` files to create and manage is up to you. But we recommend t
 | .env.test.local        | test        | no               | Local overrides of test-specific settings. |
 | .env.test              | test        | yes              | Shared test-specific settings.             |
 | .env                   | development and test | yes     | Shared settings applicable in test and development. |
+
+We do not recommend using dotenv in production environments.
+
+<p class="notice">
+  Hanami will only use the dotenv gem if it is included in your Gemfile. Applications generated using "hanami new" include the gem in development and test by default.
+</p>
--- a/content/v2.0/application/settings.md
+++ b/content/v2.0/application/settings.md
@@ -272,3 +272,64 @@ module Bookshelf
   end
 end
 ```
+
+## Using dotenv to manage environment variables
+
+Hanami uses the [dotenv gem](https://github.com/bkeepers/dotenv) to load environment variables from `.env` files.
+
+This allows you to maintain specific sets of per-environment variables for your app settings. Which set is used depends on the current environment, which is determined by `HANAMI_ENV`.
+
+`.env.development` is used if `HANAMI_ENV` is "development":
+
+```shell
+# .env.development
+DATABASE_URL=postgres://localhost:5432/bookshelf_development
+ANALYTICS_ENABLED=true
+```
+
+```shell
+HANAMI_ENV=development bundle exec hanami console
+
+bookshelf[development]> Hanami.app["settings"].database_url
+=> "postgres://localhost:5432/bookshelf_development"
+
+bookshelf[development]> Hanami.app["settings"].analytics_enabled
+=> true
+```
+
+`.env.test` is used if `HANAMI_ENV` is "test":
+
+```shell
+# .env.test
+DATABASE_URL=postgres://localhost:5432/bookshelf_test
+ANALYTICS_ENABLED=false
+```
+
+```shell
+HANAMI_ENV=test bundle exec hanami console
+
+bookshelf[test]> Hanami.app["settings"].database_url
+=> "postgres://localhost:5432/bookshelf_test"
+
+bookshelf[test]> Hanami.app["settings"].analytics_enabled
+=> false
+```
+
+For a given `HANAMI_ENV` environment, the `.env` files are looked up in the following order, which adheres to the recommendations made by the [dotenv gem](https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use):
+
+- .env.{environment}.local
+- .env.local (unless the environment is `test`)
+- .env.{environment}
+- .env
+
+This means a variable in `.env.development.local` will override a variable declared in `.env.development`.
+
+Exactly which `.env` files to create and manage is up to you. But we recommend the following as a reasonable set up for development and test environments in shared projects:
+
+| Filename               | Environment | Checked into git | Purpose                                           |
+|------------------------|-------------|------------------|---------------------------------------------------|
+| .env.development.local | development | no               | Local overrides of development-specific settings. |
+| .env.development       | development | yes              | Shared development-specific settings.             |
+| .env.test.local        | test        | no               | Local overrides of test-specific settings. |
+| .env.test              | test        | yes              | Shared test-specific settings.             |
+| .env                   | development and test | yes     | Shared settings applicable in test and development. |
--- a/content/v2.0/application/settings.md
+++ b/content/v2.0/application/settings.md
@@ -3,7 +3,7 @@ title: Settings
 order: 10
 ---
 
-Hanami supports application settings through a `Settings` class, defined in `config/settings.rb`.
+You can define your own settings for your app through a `Settings` class, defined in `config/settings.rb`.
 
 ```ruby
 # config/settings.rb
@@ -19,9 +19,11 @@ module Bookshelf
 end
 ```
 
-Using this class, you can define what settings exist within your application, what types and defaults they have, and whether or not they are required for your application to boot.
+Using this class, you can specify what settings exist within your application, what types and defaults they have, and whether or not they are required for your application to boot.
 
-Each setting is read from an environment variable matching its name. For example, with the following configuration, the Redis URL and Sentry DSN are sourced from the `REDIS_URL` and `SENTRY_DSN` environment variables.
+These "app settings" are unrelated to ["app config"](/v2.0/application/config/), which configure framework behaviours. App settings are your own to define and use.
+
+Each app setting is read from an environment variable matching its name. For example, the Redis URL and Sentry DSN settings below are sourced from `REDIS_URL` and `SENTRY_DSN` environment variables respectively.
 
 
 ```ruby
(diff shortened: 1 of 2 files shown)


content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
+---
+title: Settings
+order: 10
+---
+
+Hanami supports application settings through a `Settings` class, defined in `config/settings.rb`.
```
timriley, line 6: I have several thoughts about this very first line 😆 

First, I think it's good to establish a shorthand term for this that we can use throughout all our guides. You're close here with "application settings," but since we call our app an `Hanami::App` and it's defined in app.rb, perhaps "app settings" is better.

We'll also want to make sure we're clear in our language so this is not confused with framework configuration (i.e. `Hanami.app.config` or `SomeSlice::Slice.config`). I think calling these "app config" or "slice config" (i.e. config as distinct from _settings_) would be the way to go.

None of this is immediately applicable to this first line, apart perhaps considering "app" instead of "application" for our vernacular — I'm keen for your thoughts on this, however.

The actual thing I think we can change about this firstly line is to more strongly emphasise that these are _user defined_ settings, so perhaps something like this:

> You can define your own settings for your app through a `Settings` class, defined in `config/settings.rb`.

wdyt?

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+    # Define your app settings here, for example:
+    #
+    # setting :my_flag, default: false, constructor: Types::Params::Bool
+  end
+end
+```
+
+Using this class, you can define what settings exist within your application, what types and defaults they have, and whether or not they are required for your application to boot.
+
+Each setting is read from an environment variable matching its name. For example, with the following configuration, the Redis URL and Sentry DSN are sourced from the `REDIS_URL` and `SENTRY_DSN` environment variables.
```
timriley, line 24:
```suggestion
Each setting is read from an environment variable matching its name. For example, with the following settings, the Redis URL and Sentry DSN are sourced from the `REDIS_URL` and `SENTRY_DSN` environment variables.
```

For the same reasons I outlined in my first comment, let's stick to "settings" over "configuration" here whenever we can.

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+ANALYTICS_ENABLED=true MAX_CART_ITEMS=100 bundle exec hanami console
+
+bookshelf[development]> Hanami.app["settings"].analytics_enabled
+=> true
+
+bookshelf[development]> Hanami.app["settings"].max_cart_items
+=> 100
+```
+
+Common types that are useful in settings constructors include:
```
timriley, line 71: Around this moment it feels like the right time to mention these types come from dry-types and link to dry-types' documentation?

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+Types::Params::Bool
+Types::Params::Integer
+Types::Params::Float
+Types::Params::Date
+Types::Params::Time
+```
+
+## Required and optional settings
+
+Whether or not each setting is required for your application to start is determined by its constructor.
```
timriley, line 84:
```suggestion
Whether or not each setting is required for your application to boot is determined by its constructor.
```

Let's be consistent with "boot" instead of start?

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+
+If a setting uses a constructor like `Types::String` or `Types::Params::Bool`, then Hanami will raise an exception if the setting is missing, rather than let your application start in a potentially invalid state.
+
+The below settings configuration will result in a `Hanami::Settings::InvalidSettingsError` when `DATABASE_URL` and `ANALYTICS_ENABLED` environment variables are not set:
```
timriley, line 88:
```suggestion
The below settings will result in a `Hanami::Settings::InvalidSettingsError` when `DATABASE_URL` and `ANALYTICS_ENABLED` environment variables are not set:
```

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+! Unable to load application: Hanami::Settings::InvalidSettingsError: Could not initialize settings. The following settings were invalid:
+
+analytics_enabled:  cannot be coerced to false
+max_cart_items: can’t convert nil into Integer
+```
+
+The same exception will be raised if a setting can't be correctly coerced:
+
+```shell
+ANALYTICS_ENABLED=true MAX_CART_ITEMS="not coerceable to integer"  bundle exec hanami server
```
timriley, line 115:
```suggestion
ANALYTICS_ENABLED=true MAX_CART_ITEMS="not coerceable to integer" bundle exec hanami server
```

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+    setting :redis_url, default: "redis://localhost:6379", constructor: Types::String
+    setting :analytics_enabled, default: false, constructor: Types::Params::Bool
+    setting :max_cart_items, default: 100, constructor: Types::Params::Integer
+  end
+end
+```
+
+## Constraints
+
+To enforce additional contraints on settings, you can use a [dry-types](https://dry-rb.org/gems/dry-types) constraint in your constructor.
```
timriley, line 157: If we introduce dry-types earlier, then we can instead link directly to dry-types' constraint docs here:

```suggestion
To enforce additional contraints on settings, you can use a [constraint](https://dry-rb.org/gems/dry-types/1.2/constraints/) in your constructor type.
```

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+
+    setting :session_secret, constructor: Types::String.constrained(min_size: 32)
+    setting :from_email, constructor: Types::String.constrained(format: EMAIL_FORMAT)
+  end
+end
+```
+
+## Using settings within your application
+
+Hanami makes settings available within your application through a built-in settings provider.
```
timriley, line 178:
```suggestion
Hanami makes your settings available within your application as a `"settings"` component.
```

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+
+```ruby
+Hanami.app["settings"]
+```
+
+### Accessing settings within components
+
+To access settings within one of your components, use Hanami's Deps mixin:
```
timriley, line 186:
```suggestion
To access settings within one of your components, use the Deps mixin:
```

Do we need "Hanami's" here, since we're already in the Hanami guides? I actually don't know, keen for your thoughts.

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+  class Settings < Hanami::Settings
+    EMAIL_FORMAT = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z/i
+
+    setting :session_secret, constructor: Types::String.constrained(min_size: 32)
+    setting :from_email, constructor: Types::String.constrained(format: EMAIL_FORMAT)
+  end
+end
+```
+
+## Using settings within your application
```
timriley, line 176: Perhaps this could be "Using settings within your app and slices", and underneath the example below, we could have an example of how the settings are also available as a `"settings"` component within every slice?

The reason not to do this here is if we planned to have a separate part in the dedicated slices section that covers all the shared components from the app. TBH this might be better, because it helps the other docs like this one remain focused on their main topic.

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+    end
+  end
+end
+```
+
+For more information on components and the Deps mixin, see the [architecture guide](/v2.0/architecture/containers/).
+
+### Accessing settings within providers
+
+When registering a provider, settings can be access via the `target` method, which exposes the application's container:
```
timriley, line 210:
```suggestion
When registering a provider, you can access the app's settings via the `target`, which returns the app's container:
```

content/v2.0/application/settings.md
```diff
@@ -0,0 +1,266 @@
...
+    register "redis", redis
+  end
+end
+```
+
+See the guides for [providers](/v2.0/architecture/providers/) and [containers](/v2.0/architecture/containers/) for more information.
+
+### Accessing settings within your app class
+
+In some cases, you may want to use a setting to inform an application configuration inside your `App` class. Within the `App` class, settings are exposed at the class level for you, via a `settings` method:
```
timriley, line 232: Perhaps after the code example we could have a short note saying "Because settings can be accessed at this early point in the app's boot, it's important that the settings file is self-contained, with no dependencies on other code in your app." (Happy for you to finesse the wording)

content/v2.0/application/settings.md
timriley: At the same time it might be worth saying that this `Types` module is auto-generated for you when you subclass `Hanami::Settings`.

timriley (approved):
Fantastic introduction to settings, @andrewcroome! Some thoughts for you, feel free to include as you see best :)

content/v2.0/application/settings.md
timriley: And should probably also mention that "These types will be available as long as you keep dry-types in your Gemfile" or similar.

content/v2.0/application/settings.md
andrewcroome: Yep, excellent distinction to make clear. I've gone with your opening line, followed by:

```
Using this class, you can specify what settings exist within your application, what types and defaults they have, and whether or not they are required for your application to boot.

These "app settings" are unrelated to ["app config"](/v2.0/application/config/), which configure framework behaviours. App settings are your own to define and use.
```

content/v2.0/application/settings.md
andrewcroome: Done, thanks.

content/v2.0/application/settings.md
andrewcroome: Adjusted.

content/v2.0/application/settings.md
andrewcroome: Yeah, we don't need Hanami in this case I think. Especially with the link to the docs on it below.

content/v2.0/application/settings.md
andrewcroome: Yeah, I was of the mind that the settings per slice should go in the slices guide. Given the idea that slices are optional and opt in.

content/v2.0/application/settings.md
andrewcroome: Thanks - added a note to this effect.

PR #123, opened 2022-11-19
Add initial application booting guide.

andrewcroome:
This adds an initial guide to app booting, covering what a prepared and booted state means.

There may be more detail we can add here.

Ready for review!

### DIFF
--- a/content/v2.0/application/booting.md
+++ b/content/v2.0/application/booting.md
@@ -5,29 +5,29 @@ order: 40
 
 Hanami applications support a **prepared** state and a **booted** state.
 
-Whether your app is prepared or booted determines whether its components are **lazily loaded** on demand (prepared stated), or **eagerly loaded** up front (booted state).
+Whether your app is prepared or booted determines whether its components are **lazily loaded** on demand (for a prepared app), or **eagerly loaded** up front (for a booted app).
 
-This difference may not sound like much, but the ability of a prepared app to load _just enough_ of the app to undertake a particular task can be a huge performance advantage when it comes to testing, worker processes and other use cases where your app doesn't need every component to perform the job at hand.
+This difference may not sound like much, but the ability of a prepared app to load _just enough_ of the app to undertake a particular task can be a huge performance advantage when it comes to testing, worker processes and other use cases where your app doesn't need every component to perform the job at hand. A rake task that precompiles assets or migrates the database can execute quickly when is not delayed by loading the entire application.
 
-Conversely, a booted app is fully initialized, with every component ready to do its work. This is ideal whenever you want to incur initialization costs at boot time, such as when serving web requests.
+Conversely, a booted app is fully initialized, with every component loaded and ready to do its work. This is ideal whenever you want to incur initialization costs at boot time, such as when serving web requests.
 
 ## Hanami.prepare
 
 When you call `Hanami.prepare` (or use `require "hanami/prepare"`), Hanami will prepare your app for use.
 
-This process includes defining various Ruby constants required for your app to function, making the Ruby source files in `app/` autoloadable, and making your components (the classes you've added to the `app/` folder) lazily loadable.
+This process makes the Ruby source files in `app/` autoloadable and makes your components (from the same classes defined in `app/`) lazily loadable.
 
 This approach keeps load time to a minimum. As such, it’s the default mode in the Hanami console and when running tests.
 
 <p class="notice">
-  A prepared application is no less able to do its job than a booted one, it will just lazily load only the components it needs to perform a particular task on demand.
+  A prepared application will do everything a booted one can, it will just lazily load only the components it needs to perform a particular task on demand.
 </p>
 
 ## Hanami.boot
 
 When you call `Hanami.boot` (or use `require "hanami/boot"`) Hanami will do everything it does when `Hanami.prepare` is called, but then go further, starting each of your app's providers, while also eagerly loading all of your app's components up front.
 
-This is useful when you want to incur all initialization costs at boot time, such as when preparing an app to serve web requests.
+This is useful when you want to incur all initialization costs at boot time.
 
 Booting is the approach taken in Hanami's standard Puma setup. Thus, in Hanami's `config.ru` file you will see:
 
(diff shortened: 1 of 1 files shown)


content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Booting
+order: 40
+---
+
+Hanami applications support a **prepared** state and a **booted** state.
+
+Whether your app is prepared or booted determines whether its components are **lazily loaded** on demand (prepared stated), or **eagerly loaded** up front (booted state).
```
solnic, line 8:
```suggestion
Whether your app is prepared or booted determines whether its components are **lazily loaded** on demand (prepared state), or **eagerly loaded** up front (booted state).
```

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+
+This difference may not sound like much, but the ability of a prepared app to load _just enough_ of the app to undertake a particular task can be a huge performance advantage when it comes to testing, worker processes and other use cases where your app doesn't need every component to perform the job at hand.
```
solnic, line 10: Maybe add a sentence or expand the last one saying that a classic example is a rake task that migrates db or compile assets - it only needs a tiny portion of your app loaded and configured.

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+
+<p class="notice">
+  A prepared application is no less able to do its job than a booted one, it will just lazily load only the components it needs to perform a particular task on demand.
+</p>
+
+## Hanami.boot
+
+When you call `Hanami.boot` (or use `require "hanami/boot"`) Hanami will do everything it does when `Hanami.prepare` is called, but then go further, starting each of your app's providers, while also eagerly loading all of your app's components up front.
+
+This is useful when you want to incur all initialization costs at boot time, such as when preparing an app to serve web requests.
```
solnic, line 30: This is somewhat repeated (see line 12) - is this on purpose?

solnic (changes requested):

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+order: 40
+---
+
+Hanami applications support a **prepared** state and a **booted** state.
+
+Whether your app is prepared or booted determines whether its components are **lazily loaded** on demand (prepared stated), or **eagerly loaded** up front (booted state).
+
+This difference may not sound like much, but the ability of a prepared app to load _just enough_ of the app to undertake a particular task can be a huge performance advantage when it comes to testing, worker processes and other use cases where your app doesn't need every component to perform the job at hand.
+
+Conversely, a booted app is fully initialized, with every component ready to do its work. This is ideal whenever you want to incur initialization costs at boot time, such as when serving web requests.
```
timriley, line 12:
```suggestion
Conversely, a booted app is fully initialized, with every component loaded and ready to do its work. This is ideal whenever you want to incur initialization costs at boot time, such as when serving web requests.
```

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Hanami.prepare
+
+When you call `Hanami.prepare` (or use `require "hanami/prepare"`), Hanami will prepare your app for use.
+
+This process includes defining various Ruby constants required for your app to function, making the Ruby source files in `app/` autoloadable, and making your components (the classes you've added to the `app/` folder) lazily loadable.
```
timriley, line 18: "defining various Ruby constants required for your app to function" sounds too mysterious for my liking.

Could we make it a bit more specific (without going into too much detail)?

For example:

_tried to write an example and failed_

Perhaps we can just ditch it and instead focus on the impact on the user's own code? e.g.

```suggestion
This process makes the Ruby source files in `app/` autoloadable and makes your components (from the same classes defined in `app/`) lazily loadable.
```

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+
+This approach keeps load time to a minimum. As such, it’s the default mode in the Hanami console and when running tests.
+
+<p class="notice">
+  A prepared application is no less able to do its job than a booted one, it will just lazily load only the components it needs to perform a particular task on demand.
```
timriley, line 23: I imagine "no less able" tripping up non-native English speakers. Could we remove the double negative?

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+
+```ruby
+# frozen_string_literal: true
+
+require "hanami/boot"
+
+run Hanami.app
+```
+
+## Stepping through the boot process
```
timriley, line 42: I really love this section! Does a lot to demystify the boot process ❤️

content/v2.0/application/booting.md
```diff
@@ -0,0 +1,164 @@
...
+irb(main)> Hanami.app.prepared?
+=> true
+
+irb(main)> Hanami.app.keys
+=> ["settings", "notifications"]
+```
+
+Two components are now present: `settings` and `notifications`.
+
+`notifications` is an instance of `Dry::Monitor::Notifications` that's registered early in the prepare process to support app-wide notifications.
```
timriley, line 126: It might be worth adding a couple of words here to qualify "notifications"... maybe something like:

> to support inter-component notifications (for now mostly a framework-internal concern). 

I'm a little concerned about this object spooking people a bit, not the least of which because it adopts a name that might quite likely be wanted by app authors for their own components.

timriley (approved):
Great stuff, Andrew!!

content/v2.0/application/booting.md
andrewcroome: Thanks - added!

content/v2.0/application/booting.md
andrewcroome: Ta - removed here.

content/v2.0/application/booting.md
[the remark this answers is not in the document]
andrewcroome: Indeed!

content/v2.0/application/booting.md
andrewcroome: Yeah, that's heaps clearer, thanks.

content/v2.0/application/booting.md
andrewcroome: Very good call - removed it.

content/v2.0/application/booting.md
andrewcroome: I learnt a few things :)

content/v2.0/application/booting.md
andrewcroome: Thanks, adjusted!

PR #125, opened 2022-11-19
Add autoloading guide.

andrewcroome:
Adds an initial guide to autoloading.

### DIFF
--- a/content/v2.0/application/autoloading.md
+++ b/content/v2.0/application/autoloading.md
@@ -1,0 +1,153 @@
+---
+title: Autoloading
+order: 50
+---
+
+Hanami uses the [Zeitwerk](https://github.com/fxn/zeitwerk) code loader to support autoloading.
+
+This means that you do not need to require the classes and modules you write before you use them. Instead, your classes and modules are automatically available across your application.
+
+For example, the `Bookshelf::Entities::Book` class defined in the file `app/entities/book.rb` can be used in another class by simply using the constant `Bookshelf::Entities::Book`.
+
+
+```ruby
+# app/entities/book.rb
+
+module Bookshelf
+  module Entities
+    class Book
+      attr_reader :title, :author
+
+      def initialize(title:, author:)
+        @title = title
+        @author = author
+      end
+    end
+  end
+end
+```
+
+While this is convenient, it does mean you must adhere to Zeitwerk's expectations around file structure, in which file paths match constant paths.
+
+If `class Book` was changed to `class Novel` in the above file, the following error would be raised:
+
+```shell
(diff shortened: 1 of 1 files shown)


content/v2.0/application/autoloading.md
```diff
@@ -0,0 +1,153 @@
...
+
+While this is convenient, it does mean you must adhere to Zeitwerk's expectations around file structure, in which file paths match constant paths.
+
+If `class Book` was changed to `class Novel` in the above file, the following error would be raised:
+
+```shell
+Zeitwerk::NameError: expected file bookshelf/app/entities/book.rb to define constant Bookshelf::Entities::Book, but didn’t
+```
+
+Moving the file from `app/entities/book.rb` to `app/entities/movie.rb` would address this error.
```
timriley, line 38:
```suggestion
Moving the file from `app/entities/book.rb` to `app/entities/novel.rb` would address this error.
```

content/v2.0/application/autoloading.md
```diff
@@ -0,0 +1,153 @@
...
+module Bookshelf
+  class SlackNotifier
+    def self.notify(message)
+      # ...
+    end
+  end
+end
+```
+
+However, code placed in **other directories** within `lib/` **does** need a require statement. The  redis client below needs to be required (via `require "my_client/redis"`) when being used:
```
timriley, line 78: Perhaps it might be helpful here to note that using code from these other parts of lib/ is akin to using Ruby gems, which is why the `require` is needed.

timriley: Also, maybe to reinforce the point, don't call it `MyClient` but perhaps something that makes it seem more "external" than "mine", i.e. something named such that it could potentially be extracted into a gem. `CustomRedis` perhaps?

timriley (commented):
Looks good @andrewcroome!

Would it be worth having a note here about putting an `# auto_register: false` magic comment at the top of a file if its intended that you initialise the classes directly, rather than use them as components?

I know this isn't the container section, but anything you're accessing via the autoloading is almost guaranteed to be something you _don't_ want registered, so I think it's worth noting here.

timriley:
BTW, I don't think the note about excluding from auto-registration here needs to go into great detail (especially because there are other ways to do it, and that's beyond the scope of this page), but maybe just a short mention of the magic comment and a link out to the relevant section for learning more.

content/v2.0/application/autoloading.md
andrewcroome: Thanks, done!

content/v2.0/application/autoloading.md
andrewcroome: Oops!

andrewcroome:
Thanks @timriley. I've added a little bit on the magic comment and linked to the fuller explanation in the container and components section.

PR #129, opened 2022-11-20
docs/guides 2.0 logging

solnic:


content/v2.0/logger/usage.md
```diff
@@ -0,0 +1,79 @@
+---
+title: Usage
+order: 10
+---
+
+Hanami logger is compatible with the stdlib `Logger` but it supports structured logging by default. This means that you are encouraged to log *data*, which we refer to as *payloads* rather than plain text messages. This approach makes it much easier to process and understand your logs when running your system on production.
+
+In non-production environments, structured logs are turned into easy read text log entries, but the underlying log entries are represented as struct-like objects, even if you pass a text message to your logger.
```
andrewcroome, line 8:
```suggestion
In non-production environments, structured logs are turned into easy to read text log entries, but the underlying log entries are represented as struct-like objects, even if you pass a text message to your logger.
```

content/v2.0/logger/usage.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Basic usage
+
+To log a text entry, simply use a logger method with a name corresponding to the log level that you want to use. Let's say you want to log an entry with `INFO` level:
+
+```ruby
+bookshelf[development]> app[:logger].info "Hello World"
```
andrewcroome, line 15: We've been using string keys in the guides for `"settings"` etc - do we want to stay consistent with that here? Or do we prefer symbol keys?

andrewcroome (approved):
Nice! Couple of minor thoughts is all.

content/v2.0/logger/usage.md
solnic: @andrewcroome I like to use symbol keys for built-in and core deps that are shared everywhere. I already suggested we could follow this convention in the guides too.

PR #135, opened 2022-11-21
Tweak CSS a bit

solnic:
- Light gray background for inline code examples
- Less vibrant purple for header and buttons
- Same gray background for shell code blocks
- Reduce font size
- Make default font darker and use consistent 300 weight in both paragraphs and lists

This partly ports tweaks from #106 so see there for comparison screenshots.


timriley (approved):

solnic:
[Tweak CSS styles](https://trello.com/c/uF5tkc3K/356-tweak-css-styles)

jodosha (approved):

PR #146, opened 2022-11-21
Add actions-http-caching

swilgosz:
Resolves: https://trello.com/c/6bovam7H/310-actions-http-caching

### DIFF
--- a/content/v2.0/actions/http-caching.md
+++ b/content/v2.0/actions/http-caching.md
@@ -3,7 +3,7 @@ title: HTTP Caching
 order: 120
 ---
 
-Actions provide several features to help you take advantage of [HTTP caching][mdn-http-caching].
+Actions provide several features to help you make use of [HTTP caching][mdn-http-caching].
 
 [mdn-http-caching]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching
 
@@ -19,7 +19,7 @@ module Bookshelf
     module Books
       class Index < Bookshelf::Action
         def handle(request, response)
-          # Sets a `Cache-Control: public, max-age: 600` header
+          # Sets `Cache-Control: public, max-age: 600`
           response.cache_control :public, max_age: 600
         end
       end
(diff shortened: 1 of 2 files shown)


content/v2.0/actions/http-caching.md
```diff
@@ -0,0 +1,165 @@
...
+end
+```
+
+## Conditional requests
+
+[HTTP conditional requests][mdn-conditional-requests] are an interaction between the server and a client wherein the result of a request can be changed by comparing the related server resource with the value of a validator. These requests can be useful to validate the content of a client's cache for the given resource.
+
+For example, for a conditional `GET` request, there is a two-step workflow:
+
+1. The response to an initial request includes a “validator” in its headers, either the date of last modification for the resource (a Last-Modified response header and a subsequent If-Modified-Since request header), or an string identifying the version of the resource (an ETag  response header and a subsequent If-None-Match request header)
```
andrewcroome, line 76:
```suggestion
1. The response to an initial request includes a “validator” in its headers, either the date of last modification for the resource (a Last-Modified response header and a subsequent If-Modified-Since request header), or an string identifying the version of the resource (an ETag response header and a subsequent If-None-Match request header)
```

content/v2.0/actions/overview.md
```diff
@@ -45,7 +45,7 @@ Topics to cover:
 - response headers
 - redirects
 - halt
-- HTTP caching
+- exception handling
```
andrewcroome, line 48: Don't think you meant to add this back?

andrewcroome (approved):

content/v2.0/actions/overview.md
timriley: nope, accidental

timriley: I've removed that fully from the main branch now; just delete it here too IMO

timriley: (via the "Resolve conflicts" button — just be sure to "Squash and merge" afterwards)

PR #147, opened 2022-11-21
Add actions-testing section

swilgosz:
Resolves: https://trello.com/c/9GKXsM2x/311-actions-testing

### DIFF
--- a/content/v2.0/actions/testing.md
+++ b/content/v2.0/actions/testing.md
@@ -3,31 +3,29 @@ title: Testing
 order: 140
 ---
 
-Hanami pays a lot of attention to code testability and it offers advanced features to make our lives easier.
-The framework supports RSpec (default) and Minitest.
+Hanami pays a lot of attention to code testability and it offers advanced features to make our lives easier. The framework supports RSpec (default) and Minitest.
 
-## Unit Tests
+## Testing Actions
 
-First of all, actions can be unit tested.
-That means we can instantiate, exercise and verify expectations **directly on actions instances**.
+Actions are standalone objects with an interface that's easy to test. You can simply instantiate an action as your object under test and exercise its functionality.
 
 ```ruby
 # spec/app/actions/books/index_spec.rb
 
 RSpec.describe Bookshelf::Actions::Books::Index do
-  let(:params) { Hash[] }
+  subject(:action) do
+    Bookshelf::Actions::Books::Index.new
+  end
 
-  it "works" do
-    response = subject.call(params)
+  it "returns a successful response with empty params" do
+    response = action.call({})
     expect(response).to be_successful
   end
 end
 
 ```
 
-In the example above, `action` is an instance of `Bookshelf::Actions::Books::Index`. We can invoke `#call` on it, passing a Hash of parameters.
-The [implicit returning value](/actions/rack-integration) is a serialized Rack response.
-We're asserting that the returned response is successful (status is in `2XX` range).
+In the example above, `action` is an instance of `Bookshelf::Actions::Books::Index`. We can simply call it with an empty parameters hash. The [return value](/actions/rack-integration) is a serialized Rack response. We're asserting that the returned response is successful (status is in `2XX` range).
 
 ## Running Tests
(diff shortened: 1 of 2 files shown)


solnic:
@swilgosz thanks for working on this, I decided to take over and change it though because it was too late to do the PR review ping-pong.

content/v2.0/actions/testing.md
```diff
@@ -0,0 +1,189 @@
...
+RSpec.describe Bookshelf::Actions::Books::Show do
+  subject(:action) do
+    Bookshelf::Actions::Books::Index.new
+  end
+
+  it "returns a successful JSON response with book id" do
+    response = subject.call(id: "23", "HTTP_ACCEPT" => "application/json")
+
+    expect(response).to be_successful
+    expect(response.headers["Content-Type"]).to eq("#{ format }; charset=utf-8")
```
andrewcroome, line 69:
```suggestion
    expect(response.headers["Content-Type"]).to eq("application/json; charset=utf-8")
```

I don't see format set anywhere in the spec?

content/v2.0/actions/testing.md
```diff
@@ -0,0 +1,189 @@
...
+
+<p class="warning">
+Use test doubles only when the side effects are difficult to handle in a test environment. Remember to <strong>mock only your own interfaces</strong> and <strong>always use verified doubles</strong>.
+</p>
+
+## Testing requests
+
+Action tests are helpful for setting expectations on an action's low-level behavior. However, for many actions, testing end-to-end behavior may be more useful.
+
+For this, you can write request specs using [rack-test][rack-test], which comes included with your  Hanami app.
```
andrewcroome, line 167:
```suggestion
For this, you can write request specs using [rack-test][rack-test], which comes included with your Hanami app.
```

content/v2.0/actions/testing.md
```diff
@@ -0,0 +1,189 @@
...
+```ruby
+# spec/requests/root_spec.rb
+
+RSpec.describe "Root", type: :request do
+  it "is successful" do
+    # Find me in `config/routes.rb`
+    get "/"
+
+    expect(last_response).to be_successful
+    expect(last_response.body).to eql("Hello from Hanami")
```
andrewcroome, line 178:
```suggestion
    expect(last_response.body).to eq("Hello from Hanami")
```

content/v2.0/actions/testing.md
```diff
@@ -0,0 +1,189 @@
...
+  end
+end
+```
+
+<p class="notice">
+In many cases, you can rely on request tests and skip low-level action testing. Action tests only make sense when action logic becomes complex and you need to exercise many scenarios.
+</p>
+
+<p class="notice">
+Avoid test doubles< when writing request tests, since we want to verify that the whole stack is behaving as expected.
```
andrewcroome, line 188:
```suggestion
Avoid test doubles when writing request tests, since we want to verify that the whole stack is behaving as expected.
```

andrewcroome (approved):
Couple of minor typos - I could be doing something wrong locally when replicating the test by the sounds.

content/v2.0/actions/testing.md
```diff
@@ -0,0 +1,189 @@
...
+  subject(:action) do
+    Bookshelf::Actions::Books::Index.new
+  end
+
+  it "returns a successful JSON response with book id" do
+    response = subject.call(id: "23", "HTTP_ACCEPT" => "application/json")
+
+    expect(response).to be_successful
+    expect(response.headers["Content-Type"]).to eq("#{ format }; charset=utf-8")
+    expect(JSON.parse(response.body)).to eq("id" => "23")
```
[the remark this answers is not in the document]
timriley: Ah, I think we should just put `format :json` back in the class. let me do that and double test this.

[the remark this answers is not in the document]
timriley: Ah yes, this is because the response is a Rack::Response, and its `body` is an array. I'll just update the tests to get [0] and leave a comment.

PR #153, opened 2022-11-24
Add a comment about escaping square brackets when using zsh

gdonald:


gdonald:
Closing.  No feedback/action had caused a lack of interest on my end.

timriley:
Thanks for the nudge, @gdonald, and sorry this slipped through the cracks before. I've just applied your suggested note as a commit here: https://github.com/hanami/guides/commit/8f4740d1b8b8ae766f106af4438abc313938d6af

Thank you for your contribution!

PR #156, opened 2022-11-27
Add correct content type for examples

thedayisntgray:
The change made:
Screen Shot 2022-11-26 at 8 52 01 PM

What I'm seeing in my console when I follow the guides and run the test:
Screen Shot 2022-11-26 at 8 52 36 PM


timriley:
Thanks @thedayisntgray! This part of the guide was written while the out-of-the-box default format was still HTML, so it's great to have the properly reflect the actual current behavior :)

PR #162, opened 2022-12-14
Add a v2.0 views placeholder

rgarner:
In order to signpost a) when V1.3-style view support is coming to Hanami V2.x and to suggest what to do in the meantime, add a guides menu item for Views.

### DIFF
--- a/content/v2.0/views/_index.md
+++ b/content/v2.0/views/_index.md
@@ -1,0 +1,4 @@
+---
+title: Views
+order: 90
+---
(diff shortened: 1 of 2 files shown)


solnic:
Thanks for the PR. I think it's a good idea. I've seen multiple people being confused that there's no view layer yet and I do think we should make it clear in the guides. Having a TBD section like in this PR would help already. I would probably also add some additional text, maybe at the top of the getting started, explaining this as well.

WDYT @andrewcroome @jodosha @timriley?

timriley:
Yep, I like the idea 👍🏼

timriley:
Just pushed up a tweak to the wording. Happy for @andrewcroome to merge whenever he's ready :)

andrewcroome:
Merged. Thank you @rgarner!

PR #164, opened 2022-12-17
Consistently use $ as a prefix for shell commands

swilgosz:
Resolves: https://trello.com/c/QlZI0R5W/349-use-consistently-when-referring-to-shell-commands


swilgosz:
[Use $ consistently when referring to shell commands](https://trello.com/c/QlZI0R5W/349-use-consistently-when-referring-to-shell-commands)

PR #168, opened 2023-01-11
Update parameters.md

ceo-ooo-ooo:
I found a typo, Maybe we should add more URLs to dry-schema?


andrewcroome (approved):
Oops, good catch, thanks @ceo-ooo-ooo 🙏

PR #169, opened 2023-01-19
Fix actions testing path on example

Drowze:
Figured this is a typo, considering the rest of the document.

### DIFF
--- a/content/v2.0/actions/testing.md
+++ b/content/v2.0/actions/testing.md
@@ -12,7 +12,7 @@ The examples on this page use [RSpec](http://rspec.info), the test framework ins
 Actions are standalone objects with an interface that's easy to test. You can simply instantiate an action as your object under test and exercise its functionality.
 
 ```ruby
-# spec/app/actions/books/index_spec.rb
+# spec/actions/books/index_spec.rb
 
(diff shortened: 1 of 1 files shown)


content/v2.0/actions/testing.md
```diff
@@ -12,7 +12,7 @@ The examples on this page use [RSpec](http://rspec.info), the test framework ins
 Actions are standalone objects with an interface that's easy to test. You can simply instantiate an action as your object under test and exercise its functionality.
 
 ```ruby
-# spec/app/actions/books/index_spec.rb
+# spec/actions/books/index_spec.rb
```
swilgosz, line 15: Q: Because we can also test slices, wouldn't we want to change the actual spec folder `spec/actions -> spec/app/actions` instead of changing this path here?

andrewcroome: Looks like `bundle exec hanami generate action books.index` generates an action spec at `spec/actions/books/index_spec.rb` @swilgosz, so I think let's go with the change as suggested here by @Drowze.

andrewcroome (approved):
Thank you @Drowze 🙏

PR #170, opened 2023-02-07
fix mdn cache control link

Bounga:
For some reason, for this particular link, using brackets doesn't render the link properly


timriley:
I've just tested this on the site (https://guides.hanamirb.org/v2.1/actions/http-caching/) and this link seems to work fine. Will close this issue, but please let me know if you see any similar problems in the future!

PR #171, opened 2023-02-07
fix redirect_to example

Bounga:
example is calling `redirect_to` on `request) object where it should be `response`


parndt (approved):

timriley:
Thank you!

PR #163, opened 2022-12-17
Reduce font-size for code blocks

swilgosz:
Resolves: https://trello.com/c/R73U4Nba/362-fix-code-block-font-size-it-is-too-big

PR #173, opened 2023-03-06
Revert "Reduce font-size for code blocks"

timriley:
Reverts hanami/guides#163. This intersects poorly with #160, and makes all our code-formatted text way too tiny.

For example:

Screenshot 2023-03-06 at 9 32 40 pm


timriley:
Ahhh, so much better:

Screenshot 2023-03-06 at 9 50 07 pm

PR #176, opened 2023-05-25
correct the settings example for slice to contain a constructor key

ridget:
I noticed a minor typo on the slices settings page where the example code didn't have the `constructor` option key. 

This PR addresses that typo


timriley (approved):
Thanks @ridget!

PR #177, opened 2023-06-26
Guides 2.1 - add view docs 

andrewcroome:
A work in progress PR for the docs for Hanami view.


content/v2.0/views/views-and-exposures.md
```diff
@@ -108,3 +104,55 @@ module Bookshelf
...
+        include Deps[
+          "repositories.book_repo",
+          "repositories.author_repo"
+        ]
+
+        expose :book do |id:|
+          book_repo.get!(id)
+        end
+
+        expose :author do |book|
```
bkuhlmann: :thought_balloon: Is is possible to define multiple positional parameters for dependent exposures? I assume multiple followed by keyword parameters as needed?

content/v2.0/views/views-and-exposures.md
```diff
@@ -108,3 +104,55 @@ module Bookshelf
   end
 end
 ```
+
+### Depending on other exposures
+
+Sometimes one exposure will depend on value of another. You can depend on another exposure by naming it as a positional argument for your exposure block.
```
bkuhlmann: :name_badge: Might want to say *positional parameter* instead of *positional argument* since an argument is what you message the exposure with but when you define the exposure block itself, it's known as a parameter because it's part of the method definition.

content/v2.0/views/action-integration.md
```diff
@@ -0,0 +1,72 @@
...
+For example, this `Pages::Contact` action will render a `Pages::Contact` view:
+
+```ruby
+# app/actions/pages/contact.rb
+
+module Bookshelf
+  module Actions
+    module Pages
+      class Contact < Bookshelf::Action
+        def handle(*, response)
```
bkuhlmann, line 21: :thought_balloon: Is it possible to automatically render your view but still provide custom arguments to your view? For example, something like this:

``` ruby
def handle(*, response) = response[contact: repository.find[:id]]
```

I couldn't make this work so wanted to ask. I'm guessing the answer is no since anything beyond default behavior needs to be specific?

content/v2.0/views/input-and-exposures.md
```diff
@@ -166,3 +171,52 @@ module Bookshelf
   end
 end
 ```
+
+## Private exposures
```
bkuhlmann, line 211: :sparkle: This was nice to learn.

bkuhlmann (commented):
In the future, would it be possible to rebase multiple edits to the same file so the number of commits is reduced in order to reduce the amount of commits to read through? Some additional [food for thought](https://alchemists.io/articles/git_rebase) here.

Also, in case it helps, here's the [style guide](https://alchemists.io/articles/code_reviews#_emojis) I use for feedback. :wink: The emoji's improve readability while also conveying levels of importance which helps speed up the conversation.

content/v2.0/views/overview.md
```diff
@@ -0,0 +1,256 @@
...
+  <% best_sellers.each do |book| %>
+    <li><%= book.title %></li>
+  <% end %>
+</ul>
+
+```
+
+Exposures can also decorate the objects they provide to the view via Parts, which allow view specific behaviours to be added to objects.
+
+Views can also accept input (for example, they might take an `:id` parameter from an action), and renderer different formats (i.e. the same view can render both HTML and JSON).
```
parndt, line 182:
```suggestion
Views can also accept input (for example, they might take an `:id` parameter from an action), and render different formats (i.e. the same view can render both HTML and JSON).
```

did you mean render

timriley:
I've just pushed a slight reorganisation to this PR:

- Keep the views section in `/v2.0` as it is (as a placeholder)
- Create a new `/v2.1` directory and move all the new views guides into there
- Copy the "Getting started" guide from v2.0 into this v2.1 directory too, since this will need to be modified for view rendering.

This will allow us to continue working on the views guides in the place where they will eventually reside. If we need to modify any of the other 2.0 pages, we can bring those pages into 2.1 as required.

Once we're happy with all the new content, we can bring across the rest of the 2.0 content and then 2.1 will become the new default version for the guides.

I'm going to merge this PR now so we have a base of work to continue from.

PR #229, opened 2023-11-14
Hanami v2.1: Change ERB code snippets to sql

swilgosz:
We do not have a syntax highlighting for ERB snippets.

image

After:

image

Until we add the highlighting engine for ERB snippets, we should use `sql`

Reasoning:
1. We already do it in most of the erb snippets
2. We don't have too much SQL code in our guides, so replacing that will be quite quick in the future
3. Syntax highlighting works a bit better, correctly recognizing at least opening and closing tags for both HTML and Ruby parts of the code.

### DIFF
--- a/content/v2.1/cli-commands/commands.md
+++ b/content/v2.1/cli-commands/commands.md
@@ -1,6 +1,6 @@
 ---
 title: Commands
-order: 50
+order: 10
 ---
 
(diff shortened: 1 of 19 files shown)


timriley:
Thanks @swilgosz! I've just cherry-picked the appropriate commit into the main branch now.

PR #234, opened 2024-02-03
Update getting-started.md

dan-vernon:
Corrects start Hanami app server command in intro readme


cllns:
We need foreman to run the asset server too, so this change doesn't work.

But thanks for bringing it up! Because the distinction between `dev` and `server` is very unclear. I'll work on resolving that, starting with: https://github.com/hanami/cli/pull/144

PR #231, opened 2023-11-14
Guides v2.1: Link together view sections in the overview file

swilgosz:

PR #239, opened 2024-04-30
Add page about testing hanami-views

krzykamil:
Based on discussion https://discourse.hanamirb.org/t/testing-hanami-views/960/3

This PR should probably be merged after Sebastin's PR https://github.com/hanami/guides/pull/231 so that we could properly link this new page too

### DIFF
--- a/content/v2.1/views/testing.md
+++ b/content/v2.1/views/testing.md
@@ -1,0 +1,170 @@
+---
+title: Testing
+order: 101
+---
+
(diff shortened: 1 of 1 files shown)


timriley:
Thanks again for the effort in getting this started, @krzykamil! I've just pushed some changes to get the page into a good final shape. Along the way I realised that implemented https://github.com/hanami/hanami/pull/1357 last year, which makes testing _all aspects_ of parts (including usage of helpers, partials, and the context) easier than ever!

Merging this now. Thanks again. This wouldn't be published without your help!

PR #241, opened 2024-05-07
Fix typo of using part instead of scope

krzykamil:
While talking about scopes, the guides mention parts, instead of scopes


timriley:
Thanks @krzykamil!

PR #242, opened 2024-07-08
Prepare for v2.2 guides

timriley:
- Copy the content from v2.1/ to v2.2/
- In v2.2/ update all links from "v2.1/" to "v2.2/"
- Update the base version number in root `_index.md`
- Keep all global aliases on the v2.1 pages, since those are still the active ones


alectrico-pro (commented):
This guide is for 2.1.0. I. My you should write a brand new guide for 2.2.

PR #244, opened 2024-07-16
[2.2] Clean up Web app guide

cllns:
Some of these are simple fixes for mistakes, but others are my preferences for explaining what's happening. I know we don't want to overload the reader, but I think some info can be helpful. Feel free to push back, or only take some of thse.

I also updated the screenshot, but edited the DOM so hopefully we can use that same image for all our 2.2 releases, instead of having to make a new one with each patch release to keep it accurate.

I'll also go through the API one, but I want to get feedback on this one first.


cllns:
Some other notes I took:
* Should we change the book titles for switching from the fixtures to the DB records? Right now it's just the `, by AUTHOR` that's different, which could be hard to pick up on, since the titles are there in both, and the same.

* What do you think about renaming the feature specs to be what feature they're implementing for the user/app rather than their implementation? This might be too tied to my personal workflow but I find having distinct names helps when fuzzy searching files. Right now fuzzy-finding on `books/index` shows 6 files.
Here's what I propose:
  ```
  spec/features/books/create_spec.rb           -> spec/features/add_book_spec.rb
  spec/features/books/index_spec.rb            -> spec/features/list_books_spec.rb
  spec/features/books/index/pagination_spec.rb -> spec/features/paginate_books_spec.rb
  spec/features/books/show_spec.rb             -> spec/features/show_book_spec.rb
  ```

* Instead of using `#one` to get the book record, then introduce `#one!` right after that, what if we just jump to using `#one!` and mention that there's `#one` also. By default `#one!` will give them a better error. Especially if we take my propose changes about explaining materializing relations, that'd be a good place to talk about the difference.

Happy to implement any/all of these, just want feedback first.

PR #245, opened 2024-07-17
Update getting-started.md

alectrico-pro:
Hello, I thing that 

gem install hanami 

must be replaced with

gem install hanami --pre 

as stated on 
https://rubygems.org/gems/hanami/versions/2.2.0.beta1.

Thanks


parndt:
This was relevant, but no longer as 2.2 is released. Thanks @alectrico-pro !

PR #251, opened 2024-10-08
[Guides] Add Hanami CLI documentation for DB commands

swilgosz:
### Background.

For the CLI section, we'll need to have information about CLI commands for DB namespace.

### Design

Follow the same convention as with other CLI command docs.

### DIFF
--- a/content/v2.2/cli-commands/db.md
+++ b/content/v2.2/cli-commands/db.md
@@ -5,7 +5,7 @@ order: 110
 
 ## hanami db
 
-Hanami supports multiple database-related CLI commands, to help manage databases and their schemas for your app. You can list all of them, by adding the `--help` option.
+Hanami provides commands to manage all aspects of your database’s lifecycle. List all the commands with the `--help` option.
 
 ```shell
 $ bundle exec hanami db --help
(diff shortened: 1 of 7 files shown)


content/v2.2/cli-commands/db.md
```diff
@@ -0,0 +1,73 @@
...
+
+Now, when you'll run the command with the `--slice` option, separate db will be created for your slice.
+
+```shell
+$ bundle exec hanami db create
+
+# => database db/admin_bookshelf.sqlite created
+```
+
+TODO: `--app` option. How this is supposed to work?
```
swilgosz, line 72: @timriley It is unclear to me how to use this?

When I specify the `--slice` and `-app`, it seems to be ignored.

```shell
bundle exec hanami db create --slice=Main
=> database db/main_guides_app.sqlite created

bundle exec hanami db create --slice=Main --app
=> database db/main_guides_app.sqlite created
```

Could you please help me understand the context of this `--app` option?

alassek: `--app` would be the top-level `Hanami::App` rather than a slice. So `--app` and `--slice` are mutually-exclusive.

swilgosz: This is what I thought, but it's unclear how to use it. Not passing an option generates the db for app,, doesn't it? Or do I miss the env variable for app?

alassek: Ah, perhaps Tim intends these to be _additive_, so you can process multiple db configs at once

timriley: @swilgosz my intention is that you run these commands:

- _without_ `--app` or `--slice` options (so they work on _all_ databases)
- or with only a _single_ `--app` or `--slice` option (so you can work on a particular subset of databases)

I didn't consider both being provided. I'm pretty sure the implementation would just end up considering one of the args in that case.

(We should probably update our option handling to make that clear)

timriley:
Heads up: I'll review/adjust/merge this tomorrow!

timriley:
Tidied and significantly simplified (see https://github.com/hanami/guides/pull/251/commits/6093e441209676e04c3611c73f68262f4c8bdbc6). Thanks for putting this together, @swilgosz!

PR #264, opened 2024-11-04
Clean up database guides

cllns:
Just read through the DB docs and amazing work @alassek! I learned quite a bit and it's amazing how you distilled the complexity of ROM down and made it simple. 

Here are a bunch of small tweaks: feel free to cherry-pick ones you want to keep or make a GitHub 'suggestion' to un-do or tweak my copy even more, or disregard any changes you don't agree with.


content/v2.2/database/overview.md
```diff
@@ -137,9 +137,8 @@ module Main
...
 
-A Struct is not a permanent abstraction of a piece of data: it is a momentary projection of the data you requested. This means that instead of a User model that fills every role you need of a user, you could project user data as a Credential for authentication, a Role for authorization, a Visitor for displaying their identity on the page. Every projection can serve a specific purpose, and contain exactly the information you need and nothing more.
+A Struct is not a permanent abstraction of a piece of data: it is a momentary projection of the data you requested. This means that instead of a `User` model that fills every role you need of a user, you could project user data as a `Credential` for authentication, a `Role` for authorization, a `Visitor` for displaying their identity on the page. Every projection can serve a specific purpose, and contain exactly the information you need and nothing more. The way to map them to different object is [using `#map_to` on the Relation](https://rom-rb.org/learn/core/5.2/structs/#mapping-to-custom-objects).
```
alassek, line 142:
```suggestion
A Struct is not a permanent abstraction of a piece of data: it is a momentary projection of the data you requested. This means that instead of a `User` model that fills every role you need of a user, you could project user data as a `Credential` for authentication, a `Role` for authorization, a `Visitor` for displaying their identity on the page. Every projection can serve a specific purpose, and contain exactly the information you need and nothing more.
```

"object" should be plural here, but I think `map_to` belongs in the forthcoming Repository guide, this seems a little advanced for the Overview.

alassek (approved):
Thanks for the review @cllns! I just have one note but I like the rest of the edits.

cllns:
Agree about `#map_to` belonging in the Repositories page, not the overview :) Merging!

PR #269, opened 2024-11-23
Link to 2.2 instead of 1.3

cllns:
I had a few links go to 1.3 instead of 2.2

The other pages in this folder are fine because they don't overlap with 2.x guides at all


timriley:
Thank you!

PR #272, opened 2024-12-01
Fix hanami db command indentation

stephannv:
Small PR fixing `hanami db` command indentation. Reading the guides I was confused about the command being in a different indentation and realized that it was using tab as the indentation instead two spaces.

Screenshot 2024-12-01 at 16 03 37


cllns:
Made the same change to the (unreleased, as of now) 2.3 docs too.

PR #273, opened 2024-12-03
[docs]: Add a note about how to generate migrations

KonnorRogers:
I was looking through the migrations section and wasn't quite sure how to generate a migration. I eventually found it by running `hanami generate --help` and also noticed that the migration generation isn't documented in the Generators section either.


alassek:
Sorry for letting this languish, @KonnorRogers! It appears that we already addressed this in #274 so I'm going to close this, but I appreciate the help.

PR #274, opened 2024-12-06
Update available generator commands

stephannv:
Add component, migration, operation, relation, repo and struct commands.


timriley (approved):
Thank you! 😍

PR #275, opened 2024-12-29
Update building-a-web-app.md

megatux:
Fix type of test, feature test and not request test


megatux:
> Fix type of test, feature test and not request test

Also happens below, in pagination test

cllns:
Thank you!

PR #278, opened 2025-02-10
fix file name in introduction web app tutorial

jipiboily:
There is a small error in the introduction guide.

The _books.show_ template path is wrong; it doesn't reflect the overall tutorial and makes subsequent steps fail (as the ERB templates are in templates, not views).


timriley:
Thanks @jipiboily!

PR #280, opened 2025-02-18
Grouping actions

svoop:
This is a first shot at documenting the grouping of actions and why you might want to do this. A feature which came as a very pleasant surprise to me.

Not that easy to showcase it though, so I'm not sure whether my example makes much sense as is now, but it's maybe a starting point.

Discussion: https://discord.com/channels/1339005121303941210/1339005122059042888/1341206241569018018


svoop:
Moving the branch from fork and thus closing this PR in favor of #345

PR #281, opened 2025-02-19
Add section to clarify the lack of autoloading in the config dir

svoop:
Discussion: https://discourse.hanamirb.org/t/app-wide-rack-middleware-and-lib/1123/6

### DIFF
--- a/content/v2.2/app/autoloading.md
+++ b/content/v2.2/app/autoloading.md
@@ -127,6 +127,13 @@ end
 | lib/bookshelf/slack_notifier.rb | Bookshelf::SlackNotifier                   |
 | lib/custom_redis/client.rb          | require "custom_redis/client"<br /><br />  CustomRedis::Client |
 
+## Autoloading in the config directory
+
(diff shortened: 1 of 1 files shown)


content/v2.2/app/autoloading.md
```diff
@@ -127,6 +127,13 @@ end
 | lib/bookshelf/slack_notifier.rb | Bookshelf::SlackNotifier                   |
 | lib/custom_redis/client.rb          | require "custom_redis/client"<br /><br />  CustomRedis::Client |
 
+## Autoloading in the config directory
+
+The config directory is a particular case because some of the code is hit too early in the boot process for Zeitwerk to provide autoloading. Most notably, you have to explicitly `require` in:
+
+* `config/app.rb`
+* `config/routes.rb`
+* `config/settings.rb`
```
alassek, line 136: I think it's probably safe to say that anything in `config/` should not rely on autoloading, excluding possibly registrations, since those are lazy-loaded.

I suppose that's just framing in a negative way what you're saying positively, so I think we are in agreement.

svoop:
Thanks, @alassek !

PR #282, opened 2025-02-26
Add page for Content Security Policy

svoop:
:warning: This new page makes only sense after https://github.com/hanami/hanami/pull/1503 has been merged and released.


timriley (approved):
Thanks for doing this, @svoop!

Let's keep this one un-merged for now, I think we'll bundle your CSP work up in a 2.3 release (along with Rack 3 support), at which point we'll want to create a copy of the guides into `v2.3/` and target that for this PR.

svoop:
@timriley Doing a bit of housekeeping and apparently, you've adopted the changes as part of the 2.3 guides, so closing this now obsolete PR. Thanks a bunch!

PR #283, opened 2025-03-03
Update hanami guides with redirect cases

inouire:
As discussed in PR https://github.com/hanami/router/pull/282 , this  is an update of the guides to explain external redirects use cases


timriley (commented):
Thanks @inouire!

PR #284, opened 2025-03-04
Improve database configuration guide

cllns:
From https://github.com/hanami/hanami/issues/1505, I realized we were missing some of this info in the guides.


alassek (approved):
Thanks! That looks good to me

PR #285, opened 2025-03-07
Fix/improve config no_auto_register_paths docs

cllns:
Will extend to 2.0 and 2.1 once this review gets approved.

### DIFF
--- a/content/v2.2/app/container-and-components.md
+++ b/content/v2.2/app/container-and-components.md
@@ -324,7 +324,7 @@ end
 
 If you have a whole class of objects that shouldn't be placed in your container, you can configure your Hanami application to exclude an entire directory from auto registration by adjusting its `no_auto_register_paths` configuration.
 
-Here for example, the `app/structs` directory is excluded, meaning nothing in the `app/structs` directory will be registered with the container:
+Here for example, the `app/values` directory is excluded, meaning nothing in the `app/values` directory will be registered with the container:
 
 ```ruby
 # config/app.rb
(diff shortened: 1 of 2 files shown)


content/v2.2/app/container-and-components.md
```diff
@@ -333,11 +333,17 @@ require "hanami"
...
-    config.no_auto_register_paths << "structs"
+    config.no_auto_register_paths << "values"
   end
 end
 ```
 
+Note that the default value for `no_auto_register_paths` is `["db", "entities", "relations", "structs"]`,
+so you do not need to specify those. Also, be sure to append to that list.
+
+These apply for the root of the application, as well within any slices.
```
timriley, line 344: I think we should link to slices here. If a reader is going through the docs in rough order of appearance, they won't have got to slices yet.

timriley (approved):
Looks good. Thanks! One minor suggestion, otherwise happy to see this go in.

content/v2.2/app/container-and-components.md
cllns: Good call, added!

This makes me think we should consider re-ordering the guides, as well (perhaps on the new site). The current order is: Booting, Container and components, Providers, Settings, Autoloading, Environments, App config, Inflector, Code reloading, Slices.

Slices are more core than code reloading, or the inflector IMO. I think Autoloading and Environments could be lower as well. Perhaps: Booting, Container and components, Providers, Settings, App config, Slices, Autoloading, Environments, Inflector, Code reloading?

I still think linking to Slices is a good idea regardless of the order.

timriley: Thanks @cllns :) Agree that it'd be worth trying a different order. We came up with this order years ago for 2.0 and haven't reconsidered it since. I think an "Overview" page for the App guide would be useful thing to have too. That way we can introduce all the key concepts in relation to each other, and provide relevant jumping off links to learn more.

timriley (approved):
Perfect! Thanks!

parndt:
@cllns reminder that this is approved and ready to merge when you're happy - though it'll need to be extended to 2.3 as well now?

PR #287, opened 2025-03-09
Fix a repo class name

y-yagi:
It seems that the `DB` namespace is missing. Ref #246.


timriley:
Great pickup, thank you!

PR #288, opened 2025-03-14
Fix a path of template file

y-yagi:
The template files are under the `templates` directory in v2.2.


timriley:
Thanks @y-yagi, really appreciate these improvements! :)

PR #289, opened 2025-03-16
Add missing receiver when using `params`

y-yagi:
I think we can't use `params` without a receiver now.


timriley (approved):
You're right! Thank you as always for improving our docs 😍

PR #291, opened 2025-04-17
Make call method parameter consistent in operations docs to avoid con…

marciotoshio:
…fusion.


timriley:
Great pickup, thank you!

PR #292, opened 2025-04-21
Update building-an-api.md

igbanam:
Fixed tiny error with guides.

When the index action is generated, the guides hint it's generated into the `show` file. This change corrects that

### DIFF
--- a/content/v2.2/introduction/building-an-api.md
+++ b/content/v2.2/introduction/building-an-api.md
@@ -136,7 +136,7 @@ $ bundle exec hanami generate action home.index --skip-view --skip-route --skip-
 We can find this action in our `app` directory at `app/actions/home/index.rb`:
 
 ```ruby
-# app/actions/home/show.rb
+# app/actions/home/index.rb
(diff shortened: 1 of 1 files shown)


cllns:
Thanks! Looks like this is a clean up for #262.

Do you mind fixing lines 128 and 130 too? [They also reference "home show"](https://github.com/hanami/guides/blob/5bc867059b1aaf81711c132259dabeb7c9761019/content/v2.2/introduction/building-an-api.md?plain=1#L128-L130). They should say "home index" instead.

igbanam:
Thanks for looking through @cllns. I've updated the lines requested

cllns:
I added a commit to apply these changes to the 2.1 branch as well, and realized that the changes from #262 to 2.2 guides didn't get applied to 2.1, so I applied those as well. It was a tedious change so I figured I might as well do it myself rather than ask you to do it :)

Thanks again for your careful eye! Please open more PR's or issue for anything else you find 🌸

PR #293, opened 2025-05-23
Add guide for building a static site

svoop:
So far, the guides introduce Hanami as a framework for (dynamic) web apps and APIs. However, Hanami ❤️ Parklife make a great team to build static sites as well.

The Parklife install docs don't work for Hanami and figuring out the necessary tweaks in the generated `Parkfile` is not so trivial if you don't know where to go for a peek (e.g. the Hanami site). So why not add another intro guide to showcase that Hanami can go static, too?

Well, here it is. :smile:

### DIFF
--- a/content/v2.2/introduction/building-a-static-site.md
+++ b/content/v2.2/introduction/building-a-static-site.md
@@ -1,0 +1,50 @@
+---
+title: "Building a static site"
+order: 30
+---
+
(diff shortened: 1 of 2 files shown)


parndt:
@timriley does this match your experience with https://github.com/hanakai-rb/site ?

PR #296, opened 2025-06-20
Fix rspec metadata

KamilMilewski:
The way it is presented in documentation currently seems to not invoke db hooks specified in `spec/support/db/cleaning.rb` (noticed this when I added unique index to some db column and was getting record not unique errors from database).

### DIFF
--- a/content/v2.0/introduction/getting-started.md
+++ b/content/v2.0/introduction/getting-started.md
@@ -713,7 +713,7 @@ end
 With our books table ready to go, let's adapt our books index spec to expect an index of persisted books:
 
 ```ruby
-RSpec.describe "GET /books", type: [:request, :database] do
+RSpec.describe "GET /books", :db, type: :request do
   let(:books) { app["persistence.rom"].relations[:books] }
 
   before do
@@ -811,7 +811,7 @@ Let's add a request spec verifying pagination:
 ```ruby
 # spec/requests/books/index/pagination_spec.rb
 
-RSpec.describe "GET /books pagination", type: [:request, :database] do
+RSpec.describe "GET /books pagination", :db, type: :request do
   let(:books) { app["persistence.rom"].relations[:books] }
 
   before do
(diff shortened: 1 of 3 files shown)


KamilMilewski:
> Hi, sorry you ran into this issue and thanks for opening a PR.
> 
> I think this might be due to a misunderstanding with the versions of the docs. This PR edits Hanami 2.0 and Hanami 2.1 guides.
> 
> Hanami 2.2 introduced hanami-db and `hanami new` generates a `spec/support/db/cleaning.rb` file that uses a `:db` tag. (This happens [via `hanami/rspec`](https://github.com/hanami/rspec/pull/28)
> 
> In Hanami 2.0 and 2.1, we didn't have built-in DB support so the process was more manual. The guides instructed users to create [`spec/support/database_cleaner.rb`](https://github.com/hanami/guides/blob/main/content/v2.0/introduction/getting-started.md?plain=1#L603-L621), which used the `type: database` (vs the `:db` tag as we do in Hanami 2.2)
> 
> My guess is that you were on the 2.0 or 2.1 docs but using Hanami 2.2. Is that right?
> 
> Therefore, I think this PR is invalid since `:db` tag doesn't exist in 2.0 and 2.1. Do you agree? If so can you close this? But don't hesitate to open PR's in the future :)

Hey @cllns , thanks for taking time into looking into this. I was actually following 2.2 docs but made a mistake in this PR and edited 2.0 & 2.1 instead 🙈 , sorry for the confusion!
So the original issue is still valid I think, it's just not manifesting itself when following `2.2 getting started guides` because specs there don't rely on database cleaning between examples.
Other option that seems to work would be to change generated spec/support/db/cleaning.rb file, so that we would have:
```
config.before :each, type: :db
```
instead of:
```
config.before :each, :db do |example|
```
(same for `after` hook)

cllns:
Aha! Yes, I see. We want to use the `:db` tag over the `type:` so the changes look right now!

We don't need to update the 'building a web app guide' because the `spec/support/db.rb` file automatically adds the `:db` tag to all `:feature` specs. We don't use feature specs, but rather `:request` specs so that doesn't apply.

Thanks again!

PR #298, opened 2025-06-21
A suggested add to views/parts and a suggested fix to views/scopes.

blazerw:
I was confused by the Memoization section of Views/Parts because it just seemed to repeat the example from Decorating above. But, it appears, correct me if I'm wrong, that `decorate` memoizes as well. I added some text to make that clear.

InViews/Scopes, "exposure" seemed wrong. I think it should be "scope".

### DIFF
--- a/content/v2.2/views/parts.md
+++ b/content/v2.2/views/parts.md
@@ -202,7 +202,7 @@ decorate :author, as: :person
 
 ## Memoizing methods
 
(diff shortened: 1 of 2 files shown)


fz-ryanbigg (approved):

timriley:
Thanks you, @blazerw, this is very helpful!

PR #299, opened 2025-08-04
Provide more accurate info about running hanami in prod

krzykamil:
Based on this https://discourse.hanamirb.org/t/starting-hanami-in-production-environment/1273/2 we should make it more clear that due to the hanami server being just a thin wrapper on top of Rack::Server running the `hanami server` in production is not really supported and instead a more direct approach, like running puma with config file, is recommended.

Also removed the mentioned of RACK_ENV as recommended variable to use as it  was only ever meant to define the default middleware.

Changed the order of the text too, to make it more explicit and visible, as I consider running the app in production env to be a crucial information.

### DIFF
--- a/content/v2.2/app/environments.md
+++ b/content/v2.2/app/environments.md
@@ -7,9 +7,23 @@ Hanami supports different environments based on a `HANAMI_ENV` environment varia
 
 Setting `HANAMI_ENV` allows your code to act differently, depending on the environment.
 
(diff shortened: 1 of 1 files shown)


timriley:
Thanks @krzykamil :) As it turns out, I'm not so certain we actually need these changes. I left some extra thoughts [on the original forum thread](https://discourse.hanamirb.org/t/starting-hanami-in-production-environment/1273/9).

krzykamil:
Closing in favor of https://github.com/hanami/reloader/pull/32

PR #300, opened 2025-08-04
Add note about --slice option being omitable due to cwd

krzykamil:
Adding a little note due to https://github.com/hanami/cli/pull/298


krzykamil:
> This is great, thanks @krzykamil! I think this will actually release as part of 2.3, though, so let's hold back this PR and merge it once 2.3 docs are ready.
> 
> (In fact, if you have time would you like to do that in another PR, that would be very helpful! It's just a matter of copying the 2.2 folder to 2.3).

https://github.com/hanami/guides/pull/302

PR #303, opened 2025-09-18
Exclude v2.3 from dropdown

cllns:
Clean up for #302.

It's fine if the pages are around, IMO, but we don't want people to discover them. I also added a robots.txt to remove the already-indexed pages from search engines, which I'll be responsible for removing once we launch 2.3


timriley (approved):
Thanks Sean :)

PR #306, opened 2025-10-11
Update action format docs for new 2.3 API

timriley:
Update the "Formats and MIME types" docs to use the new config API we're introducing in https://github.com/hanami/controller/pull/485

Also add the beginnings for a 2.3 upgrade guide, covering the action format changes.

### DIFF
--- a/content/v2.3/upgrade-notes/v2.3.md
+++ b/content/v2.3/upgrade-notes/v2.3.md
@@ -1,0 +1,57 @@
+---
+title: Upgrade to 2.3
+order: 30
+---
+
+These notes cover an upgrade from 2.2 to 2.3.
+
+## Use new formats config for actions
+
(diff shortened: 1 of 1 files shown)


content/v2.3/actions/formats-and-mime-types.md
```diff
@@ -7,35 +7,36 @@ Hanami maps [over 50 of the most common MIME types][built-in-formats] to simple
...
-config.actions.format :json, :html
+config.actions.formats.accept :json, :html
 ```
 
-## Configuring a format for particular actions
+## Accepting a format for particular actions
 
-You can also configure a format on any action class. `format` in an action class is analogous to `config.actions.format` in your app class, just as `config` in an action is analogous to `config.actions` in your app.
+You can also accept formats on any individual action class. `config.formats` in an action class is
+analogous to `config.actions.formats` in your app class.
```
katafrakt: Just a random thought, but maybe "works the same way as" is more precise than "is analogous to", which suggests similarity with differences (at least to this non-native English speaker)

katafrakt (commented):
few minor nitpicks

PR #326, opened 2025-11-08
Add 2.3 upgrade notes

### LINKED ISSUE #313: Write 2.3 upgrade notes

timriley:
Resolves #313

### DIFF
--- a/content/v2.3/upgrade-notes/v2.3.md
+++ b/content/v2.3/upgrade-notes/v2.3.md
@@ -19,9 +19,19 @@ Since Hanami 2.3 allows Rack version 3 for the first time, you should specify yo
 gem "rack" "~> 2.2"
 ```
 
+## Disable Foreman’s own env parsing
+
+Update your `bin/dev` to start `foreman` like so:
+
+```shell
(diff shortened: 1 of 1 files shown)


content/v2.3/upgrade-notes/v2.3.md
```diff
@@ -0,0 +1,178 @@
+---
+title: Upgrade to 2.3
+order: 10
+---
+
+These notes cover an upgrade from 2.2 to 2.3.
+
+## Update the app
+
+- Edit `Gemfile` and update the versions of the Hanami gems to `"~> 2.3"`.
```
parndt: are we sure we don't want to suggest `"~> 2.3.0"` so that they're not automatically updated to `[>= 2.3, < 3.0]`, or is that what we want?

timriley: Good point!

parndt:
```suggestion
- Edit `Gemfile` and update the versions of the Hanami gems to `"~> 2.3.0"`.
```

